PR #7, opened 2019-02-16
Billy Herrington playermodel + SWEP

chev2:
Adds Billy Herrington temp playermodel and a SWEP similar to squee/deusvult/atheist that is equipped upon buying the model.


swampservers:
looks good, thanks!

PR #12, opened 2019-02-23
model reorganization, new slot machine texture, and slitter killicon

Noztik:
I need the following code to be added to the slitter for it to have the correct css knife killicons since the slitter code isn't on the repo
`if (CLIENT) then
	killicon.AddFont("weapon_slitter", "CSKillIcons", SWEP.IconLetter, Color( 255, 80, 0, 255 ))
	surface.CreateFont("CSKillIcons", {font = "csd", size = ScreenScale(30), weight = 500, antialias = true, additive = true})
	surface.CreateFont("CSSelectIcons", {font = "csd", size = ScreenScale(60), weight = 500, antialias = true, additive = true})
end`


swampservers:
found issues

materials/models/noz/partyhat.vmt
```diff
@@ -1,4 +1,4 @@
-"VertexlitGeneric"
-{
-	"$basetexture" "models/misc/PartyHat.vtf"
+"VertexlitGeneric"
+{
+	"$basetexture" "models/misc/PartyHat.vtf"
```
swampservers: need to rename this to point to the moved texture

Noztik: I had forgotten to change all vmt data for all the materials

PR #15, opened 2019-03-06
gmt_instrument sound compression

chev2:
Compresses all the .wav instrument files from gmodtower/lobby/instruments/ (which took up about ~167MB in total) into an .ogg format (which takes up around ~15MB in total). this should make it easier for newfriends from third world countries to download the server content with their 128kbps internet speeds


swampservers:
Excellent!

PR #16, opened 2019-03-09
Delete unnecessary and ugli normal map

Milwacko:


swampservers:
thanks, can someone also lower the triangle count of this?

PR #18, opened 2019-03-10
Pepsiman model

chev2:
DRINK!


swampservers:
Those .vtfs are MASSIVE, a lot of workshop addons are made by idiots who only use 4096x4096 uncompressed. Can you make them 512x512 DXT1s?

chev2:
ok, I decreased the filesize of the 10 mb files significantly, so it should take up a lot less room now. if the files are still too big, ill try resizing them again

edit: they are currently at 1024x512 because the files aren't squares

swampservers:
Thanks, sorry for delay

PR #19, opened 2019-03-13
pit respawn location and minor additions

Noztik:


swampservers:
Are you sure about this? The spawn is already right next to the pit. This will make people spawn in the same place each time too

chev2:
You can either do a math.random function in that place where you put the spawnpoint vector to spread players out more evenly (to prevent spawncamping on respawn), or alternatively (and I don’t know if this’ll work), find all info_player_start entities exclusive to the lobby location and just spawn players to those entity locations

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Noztik:
people just didn't want to spawn far away from the pit when they died while fitting in the pit

also the "finding all the info_player_start entities" idea should be possible and is a good idea

lua/autorun/misc.lua
```diff
@@ -67,4 +67,29 @@ else
 			net.SendToServer()
 		end
 	end
+	
+	hook.Add("PlayerDeath","DeathInPit",function(ply)
+		if IsValid(ply) then
```
swampservers: this hook would only be called with a valid player. unnecessary checks

lua/autorun/misc.lua
```diff
@@ -67,4 +67,29 @@ else
 			net.SendToServer()
 		end
 	end
+	
+	hook.Add("PlayerDeath","DeathInPit",function(ply)
+		if IsValid(ply) then
+			local loc = ply:IsPlayer() and ply:GetLocation() or Location.Find(ply)
+			local name = Location.GetLocationNameByIndex(loc)
+			if name == "The Pit" then
```
swampservers: if ply:GetLocationName() == "The Pit" then

lua/autorun/misc.lua
```diff
@@ -67,4 +67,29 @@ else
...
+		if IsValid(ply) then
+			local loc = ply:IsPlayer() and ply:GetLocation() or Location.Find(ply)
+			local name = Location.GetLocationNameByIndex(loc)
+			if name == "The Pit" then
+				ply.PitDeath = true
+			end
+		end
+	end)
+
+	hook.Add("PlayerSpawn","SpawnNextToPit",function(ply)
```
swampservers: The proper way to do this is with https://wiki.garrysmod.com/page/GM/PlayerSelectSpawn

lua/entities/gmt_instrument_base/shared.lua
```diff
@@ -64,7 +64,14 @@ function ENT:GetSound( snd )
 		return nil
 	end
 	
-	return "gmodtower/lobby/instruments/" .. self.Instruments[self:GetNWInt("CurrentInstrument")] .. "/" .. snd .. ".ogg"
```
swampservers, line 67: You can do:
return "gmodtower/lobby/instruments/" .. (self.Instruments[self:GetNWInt("CurrentInstrument")] or "Piano") .. "/" .. snd .. ".ogg"

swampservers (changes requested):
commnents, sorry for delay was on vacation

lua/autorun/misc.lua
```diff
@@ -69,27 +69,21 @@ else
...
-		if IsValid(ply) then
-			if ply.PitDeath then
-				local outsidespawns = {}
-				for k,v in pairs(ents.FindByClass("info_player_start")) do
-					if (v:GetPos().y == -64 or v:GetPos().y == -160 or v:GetPos().y == -256) then
-						table.insert(outsidespawns,v:GetPos())
-					end
+	hook.Add("PlayerSelectSpawn","SpawnNextToPit",function(ply)
+		if ply.PitDeath then
+			local outsidespawns = {}
```
swampservers: Generating this table is expensive, cache it. Do a global variable like OutsideSpawnPoints and if it is nil (the default) then build the table

lua/autorun/misc.lua
```diff
@@ -69,27 +69,21 @@ else
...
+			for k,v in pairs(ents.FindByClass("info_player_start")) do
+				if (v:GetPos().y == -64 or v:GetPos().y == -160 or v:GetPos().y == -256) then
```
swampservers: Did you test this? There might be floating point error, better to just do v:GetPos().y < - 63 and v:GetPos().y > -257

lua/autorun/misc.lua
```diff
@@ -69,27 +69,21 @@ else
...
+					table.insert(outsidespawns,v)
 				end
-				ply:SetPos(outsidespawns[math.random(1,#outsidespawns)])
-				ply.PitDeath = false
 			end
+			ply.PitDeath = false
+			return outsidespawns[math.random(#outsidespawns)]
```
swampservers: return nil if the table is empty, in case we switch map or whatever

swampservers (changes requested):
comments

lua/autorun/misc.lua
swampservers: It would also be prudent to regenerate the table if the value you hit is not IsValid(), in case it was deleted

lua/autorun/misc.lua
Noztik: I tested it and each y unit worked fine but I'll still make the change

lua/autorun/misc.lua
```diff
@@ -76,14 +77,22 @@ else
 
 	hook.Add("PlayerSelectSpawn","SpawnNextToPit",function(ply)
 		if ply.PitDeath then
-			local outsidespawns = {}
-			for k,v in pairs(ents.FindByClass("info_player_start")) do
-				if (v:GetPos().y == -64 or v:GetPos().y == -160 or v:GetPos().y == -256) then
-					table.insert(outsidespawns,v)
+			ply.PitDeath = false
+			if IsValid(outsidespawns) and #outsidespawns < 1 then
```
swampservers, line 81: IsValid is really more for entities. If you want to check that it is not nil just do if outsidespawns...

swampservers:
I'd use outsidespawns=nil for when it's not set up but good enough

PR #26, opened 2019-05-02
Small bug fixes

chev2:
fixing some stuff


swampservers:
Can't merge, it says "draft"

chev2:
I'll set it to ready for review after I push the other commits

chev2:
The reason the pit respawn wasn't working was because IsValid(outsidespawns) was being called on an empty table, so I removed that valid check. If it absolutely _has_ to be validated, then someone else is going to have to figure out that part of the lua since I don't know how to do it myself @NottyNoz

swampservers:
You're not actually supposed to IsValid the table, afaik it's only meant to be used on entities

Great additions! We can probably give you staff if you'd like

PR #28, opened 2019-06-01
Anti-Pony Rifle

chev2:
Let me know if I should make any changes so I can get them fixed ASAP

muh sexy hot pony rp

### DIFF
--- a/lua/sps/categories.lua
+++ b/lua/sps/categories.lua
@@ -42,6 +42,7 @@ PS_Categories = {
 					"weapon_shotgun",
 					"weapon_slitter",
 					"weapon_ar2",
+					"weapon_antiponyrifle",
 					"weapon_357",
 					"weapon_smg1",
 					"weapon_gauntlet",
--- a/lua/sps/items/weapons.lua
+++ b/lua/sps/items/weapons.lua
@@ -114,6 +114,14 @@ PS_WeaponAndAmmoProduct({
 })
 
 PS_WeaponAndAmmoProduct({
+	name = 'Anti-Pony Rifle',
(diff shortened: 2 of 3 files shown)


swampservers:
So it's 30 pony kills for 2000 just like the kleiner one? Sorry, that's a no.

The kleiner gun is there because he's the default model and people change their model once they figure the server out. With pony models it's the opposite, it costs money to get and no one will want it if it makes them die constantly. Everyone who already has it would be very upset by this.

chev2:
Would it be wise if I upped the price a fair amount + drastically decreased the clip size, or is it a no-go altogether? I see where you're coming from, and 30 kills for 2000 points is definitely unfair. I can't think of any other handicaps besides those, except maybe restricting the weapon use to specific areas or giving it a large cooldown between purchases. It would make sense if there was some easy, permanent guard to nullify the weapon (similar to how changing your playermodel makes you immune to the kleiner rifle) but I can't think of anything that would work for the pony model.

swampservers:
If it's at all more effective (price/kill or just power in general) than the other weapons, that would carry the same problem as it makes the game unfair against those who saved up for the model.

For this to work, there would need to be an anti-human item of roughly equal power. It might be cool if only humans could use the anti-pony rifle and vice versa. The two special weapons could be more effective and marginally better price/kill than the normal weapons, but definitely not 30 kills for 2000 as that just turns the server into constant RDM.

Maybe for the anti-human weapon it could be "pony magic" or something

chev2:
I'm not sure how I could balance this, since I don't know how many active humans are on the server compared to active ponies on average. It sounds like a pretty cool idea but I'm not sure what would need to get done in order to make sure that both weapons are balanced, and don't overpower other weapons for around the same price range. I'll close the PR, and if someone else wants to make a balanced anti-pony/anti-human weapon then they're free to work off of this.

PR #34, opened 2019-06-24
Killicons for Sniper, Jihad & Big Bomb

chev2:
Adds killicons for the above weapons. I couldn't find good icons for the two bombs in the CS:S killicon font, so I made my own.
Killicons


swampservers:
Thank you, chev, very cool!

PR #35, opened 2019-06-24
Update golf lua

chev2:
Changed where most of the SWEP properties are located, and it _should_ now make the golf club go into your inventory like normal but not autoequip it (so if you have another weapon equipped, it won't get in the way)

### DIFF
--- a/lua/weapons/weapon_golfclub/shared.lua
+++ b/lua/weapons/weapon_golfclub/shared.lua
@@ -106,15 +106,40 @@ end
 
 
 	SWEP.PrintName			= "Golf Club"
-	SWEP.Author				= "Swamp & PYROTEKNIK"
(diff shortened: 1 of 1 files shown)


swampservers:
I don't rely on AutoSwitchTo, I would manually select the weapon. The reason is so if some kleiner has the fedora on and walks into golf he doesn't get confused about how to play it because he doesn't realize that he got the club

chev2:
It might be possible to display a message on the bottom of the screen on weapon pickup -- similar to when you buy something from the shop -- saying something like "A golf club has been added to your inventory".

I wanted to make the golf club a regional pickup item exactly like the pickaxe, but I couldn't find the lua for the current autoequipping the golfclub, so I just went with the AutoSwitchTo fix.

PR #38, opened 2019-07-10
Bounty System and Golf Monthly Reset

Noztik:
you need to add the limiter to the bounty system like you have for givepoints

also it's up to you if you want to add a throttle to the chat command because the RegisterChatCommand function isn't suited for chat commands like this


lua/autorun/client/hell.lua
```diff
@@ -240,13 +240,13 @@ hook.Add("PostDrawOpaqueRenderables","HellRender",function(depth,skybox)
 	
 end)
```
swampservers: Probably revert these because of the other change

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,89 @@
...
+		elseif table.Count(search) > 1 then
+			return "Multiple players found, be more specific."
+		elseif table.Count(search) == 0 then
+			return "No players found."
+		end
+		return "Error"
+	end
+end
+
+hook.Add("PlayerSay","BountyChatCommands",function(ply,text)
```
swampservers: Can you set bounties with either a chat command or a derma window? Please do just one or the other

lua/autorun/client/bounties.lua
```diff
@@ -0,0 +1,131 @@
...
+	l1:DockMargin(4, 0, 4, 4)
+	l1:SizeToContents()
+
+	local pselect = vgui.Create("DComboBox", bounty)
+	pselect:SetValue("Select A Player")
+	pselect:SetTall(24)
+	pselect:Dock(TOP)
+
+	for _, ply in pairs(player.GetHumans()) do
+		pselect:AddChoice(ply:Nick(),ply:UniqueID())
```
swampservers: How is the UniqueID used? You can probably just net.WriteEntity the target player

lua/weapons/weapon_golfclub/shared.lua
```diff
@@ -211,268 +208,258 @@ function SWEP:SetupDataTables()
...
+			timer.Simple(0.4, function() sound.Play("golfmines/amazing.wav", vd, 70, 100, 1) end)
+			self.CantPlaceBall = true
+			timer.Simple(7,function() if IsValid(self) then self.CantPlaceBall=false end end)
+		end
+		local record = 10
+		if SERVER then
+			record = tonumber(self:GetOwner():GetPData("golf_"..hole, 10))
+			local monthlyreset = os.date("*t",tonumber(self:GetOwner():GetPData("monthly_reset",0)))
+			local curdate = os.date("*t",os.time())
+			if curdate["year"] > monthlyreset["year"] or curdate["month"] > monthlyreset["month"] then
```
swampservers: You could just have a month counter equal to year*12 + month

lua/weapons/weapon_golfclub/shared.lua
```diff
@@ -211,268 +208,258 @@ function SWEP:SetupDataTables()
...
+				for h,_ in pairs(GolfPrizeTargets) do
+					self:GetOwner():SetPData("golf_"..h,10)
+				end
+				self:GetOwner():SetPData("monthly_reset",os.time())
```
swampservers: Then store the month counter here instead of the time

lua/weapons/weapon_golfclub/shared.lua
```diff
@@ -211,268 +208,258 @@ function SWEP:SetupDataTables()
...
+			timer.Simple(7,function() if IsValid(self) then self.CantPlaceBall=false end end)
+		end
+		local record = 10
+		if SERVER then
+			record = tonumber(self:GetOwner():GetPData("golf_"..hole, 10))
+			local monthlyreset = os.date("*t",tonumber(self:GetOwner():GetPData("monthly_reset",0)))
+			local curdate = os.date("*t",os.time())
+			if curdate["year"] > monthlyreset["year"] or curdate["month"] > monthlyreset["month"] then
+				for h,_ in pairs(GolfPrizeTargets) do
+					self:GetOwner():SetPData("golf_"..h,10)
```
swampservers: Can we keep the all-time best scores and make a seperate monthly best score tracker?

swampservers (commented):
These are awesome changes, I'll definitely paypal you when they get merged

swampservers:
BTW the way the limiter works is it basically creates a global table of steamid1->steamid2->points sent, which gets cleared by the server resetting every day, its pretty simple

lua/autorun/server/bounties.lua
Noztik: why can't it be both

lua/autorun/client/bounties.lua
Noztik: it may just be a leftover from previous versions of the code

lua/autorun/server/bounties.lua
swampservers: Hard to debug/maintain. If you still want to, please make a server side function AddBountyToPlayer(calling_ply, target_ply, amount) and handle as much of the logic as possible within it so there is minimal duplicate code

Noztik: I kind of need the derma window so that people can see who all has a bounty

Noztik: I could just get rid of the whole NewBounty function but the list with the derma is all pretty necessary

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,87 @@
...
+util.AddNetworkString("IncreaseBounty")
+
+hook.Add("PlayerDeath","PlayerDeath",function(ply,infl,atk)
+	local bounty = tonumber(ply:GetPData("bounty",0))
+	if bounty > 0 and ply != atk then
+		atk:PS_GivePoints(bounty)
+		BotSayGlobal("[edgy]"..atk:Nick().." [fbc]has claimed [gold]"..ply:Nick().."'s [fbc]bounty of [rainbow]"..bounty.." [fbc]points!")
+		ply:SetPData("bounty",0)
+	end
+end)
```
chev2, line 12: I apologize if I'm missing something here, but are you checking to make sure that the attacker is a player? Just in case the player is killed by an entity, (like keemstar) you don't want to attempt to give the entity points

chev2: Referring to the PlayerDeath hook

Noztik: you're right, I didn't read the hook's info close enough

NottyNoz:
> 
> 
> BTW the way the limiter works is it basically creates a global table of steamid1->steamid2->points sent, which gets cleared by the server resetting every day, its pretty simple

Noztik:
> 
> 
> BTW the way the limiter works is it basically creates a global table of steamid1->steamid2->points sent, which gets cleared by the server resetting every day, its pretty simple

that sounds like it only limits how much you can give to a single person in one day and not how much you can give in total in a day
also should the bounty limiter count towards the same amount as the givepoints limiter or should it be it's own thing

lua/autorun/server/bounties.lua
swampservers: You could make a chatcommand to print the list of people with bounties (only those with >0 bounty)

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,87 @@
...
+			BotSayGlobal("[fbc]"..to:Nick().."'s bounty is now [rainbow]"..bounty + p.." [fbc]points")
+		else
+			ply:ChatPrint("You have increased "..target:Nick().."'s bounty by "..amount.." points",ply)
+		end
+	else
+		ply:ChatPrint("[red]You don't have enough points")
+	end
+end
+
+hook.Add("PlayerSay","BountyChatCommands",function(ply,text)
```
swampservers: Please use the chatcommand api (see /lua/chatcommands)

Noztik: what are the args that are passed to the function of RegisterChatCommand

swampservers: the command name then the callback with player and arg (the rest of the message after the command)

lua/weapons/weapon_golfclub/shared.lua
```diff
@@ -211,268 +208,262 @@ function SWEP:SetupDataTables()
...
+		if joke then
+			local vd = self.Owner:GetPos()
+			timer.Simple(0.4, function() sound.Play("golfmines/amazing.wav", vd, 70, 100, 1) end)
+			self.CantPlaceBall = true
+			timer.Simple(7,function() if IsValid(self) then self.CantPlaceBall=false end end)
+		end
+		local record = 10
+		if SERVER then
+			record = tonumber(self:GetOwner():GetPData("golf_"..hole, 10))
+			monthlyrecord = tonumber(self:GetOwner():GetPData("monthly_golf_"..hole, 10))
```
swampservers, line 378: Can you tell the player something about what the player's all-time record is vs their monthly record if the two differ?

lua/autorun/server/bounties.lua
NottyNoz: is arg a table or just a string and if it's a string would it be like "bounty noz 1000"

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,78 @@
+
+hook.Add("PlayerDeath","PlayerDeath",function(ply,infl,atk)
+	local bounty = tonumber(ply:GetPData("bounty",0))
+	if bounty > 0 and ply != atk and type(atk) == "Player" then
+		atk:PS_GivePoints(bounty)
```
swampservers, line 9: Set the bounty to zero before giving points. If there was some bug where BotSayGlobal caused an error this would be an infinite points exploit

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,78 @@
...
+		elseif table.Count(search) > 1 then
+			return "Multiple players found, be more specific."
+		elseif table.Count(search) == 0 then
+			return "No players found."
+		end
+		return "Error"
+	end
+end
+
+function AddBounty(ply,target,amount,global)
```
swampservers: Why is global optional?

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,78 @@
...
+		end
+	else
+		ply:ChatPrint("[fbc]!bounty player points")
+	end
+end, {global=true, throttle=true})
+
+RegisterChatCommand({'bounties'}, function(ply, arg)
+	local t = {}
+	for k,v in pairs(player.GetHumans()) do
+		local bounty = tonumber(v:GetPData("bounty",0))
```
swampservers: a problem here is that GetPData does a sql lookup which is very slow to do 100 times. Make wrapper functions GetPlayerBounty and SetPlayerBounty which internally cache the player's bounty value in the entity table after it has been loaded

lua/weapons/weapon_golfclub/shared.lua
```diff
@@ -211,268 +208,269 @@ function SWEP:SetupDataTables()
...
+				if strk==1 then strokes = " stroke" end
+				self:GetOwner():PrintMessage( HUD_PRINTTALK, "[red]You beat your all-time record! New record: [orange]"..tostring(strk)..strokes )
+			else
+				if record==1 then strokes = " stroke" end
+				self:GetOwner():PrintMessage( HUD_PRINTTALK, "[red]Your all-time record for this hole is: [orange]"..tostring(record)..strokes )
+			end
+			strokes = " strokes"
+			if strk<monthlyrecord then
+				if strk==1 then strokes = " stroke" end
+				self:GetOwner():PrintMessage( HUD_PRINTTALK, "[red]You beat your monthly record! New record: [orange]"..tostring(strk)..strokes )
```
swampservers: If you beat the all-time record, you don't need to repeat yourself with the monthly record

lua/weapons/weapon_golfclub/shared.lua
```diff
@@ -211,268 +208,269 @@ function SWEP:SetupDataTables()
...
+								self:GetOwner():PrintMessage( HUD_PRINTTALK, "[red]Congratulations! You won [gold]"..tostring(v[2]).." points!" )
+								self:GetOwner():PS_GivePoints(v[2])
+							end
+							iwonsomething = true
+						end
+					end
+				end
+			else
+				if monthlyrecord==1 then strokes = " stroke" end
+				self:GetOwner():PrintMessage( HUD_PRINTTALK, "[red]Your monthly record for this hole is: [orange]"..tostring(monthlyrecord)..strokes )
```
swampservers: print your all time record below this line, rather than the first thing it prints

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,78 @@
...
+hook.Add("PlayerDeath","PlayerDeath",function(ply,infl,atk)
+	local bounty = tonumber(ply:GetPData("bounty",0))
+	if bounty > 0 and ply != atk and type(atk) == "Player" then
+		atk:PS_GivePoints(bounty)
+		BotSayGlobal("[edgy]"..atk:Nick().." [fbc]has claimed [gold]"..ply:Nick().."'s [fbc]bounty of [rainbow]"..bounty.." [fbc]points!")
+		ply:SetPData("bounty",0)
+	end
+end)
+
+function FindPlayer(plyname)
```
swampservers: Use my utility function instead of this: `PlyCount(name) -> player, count`
player is one player that matches, or nil if none matches. count is the total number of matches so if count>1 then you can err

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,78 @@
...
+		ply:ChatPrint("[red]You don't have enough points")
+	end
+end
+
+RegisterChatCommand({'bounty'}, function(ply, arg)
+	local t = string.Explode(" ",arg)
+	local to = FindPlayer(t[1])
+	if type(to) == "Player" then
+		local bounty = tonumber(to:GetPData("bounty",0))
+		if tonumber(t[2]) then
```
swampservers: bring back a minimum bounty value of 1000 please

swampservers (changes requested):
commenmts

lua/autorun/server/bounties.lua
Noztik: bit from old code

lua/autorun/server/bounties.lua
Noztik: in the player's entity table?

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
...
+	local bounty = GetPlayerBounty(ply)
+	if bounty > 0 and ply != atk and type(atk) == "Player" then
+		SetPlayerBounty(ply,0)
+		atk:PS_GivePoints(bounty)
+		BotSayGlobal("[edgy]"..atk:Nick().." [fbc]has claimed [gold]"..ply:Nick().."'s [fbc]bounty of [rainbow]"..bounty.." [fbc]points!")
+	end
+end)
+
+function GetPlayerBounty(ply)
+	if not IsValid(ply.bounty) then
```
swampservers: IsValid is only for entities. just do ply.bounty==nil

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
+
+hook.Add("PlayerDeath","BountyDeath",function(ply,infl,atk)
+	local bounty = GetPlayerBounty(ply)
+	if bounty > 0 and ply != atk and type(atk) == "Player" then
```
swampservers: atk:IsPlayer()

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
...
+		ply:PS_TakePoints(amount)
+		BotSayGlobal("[fbc]"..target:Nick().."'s bounty is now [rainbow]"..GetPlayerBounty(target) + amount.." [fbc]points")
+	else
+		ply:ChatPrint("[red]You don't have enough points")
+	end
+end
+
+RegisterChatCommand({'bounty'},function(ply,arg)
+	local t = string.Explode(" ",arg)
+	local to = PlyCount(t[1])
```
swampservers: it doesn't return a table, it returns a vararg with two members. so just do ply,count = PlyCount(...)

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
...
+	if type(to[1]) == "Player" and not to[2]>1 then
+		local bounty = GetPlayerBounty(to)
+		local p = tonumber(t[2])
+		if type(p) == "number" then
+			if p > 1000 then
```
swampservers: greater than or equal to

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
...
+				AddBounty(ply,to,p)
+			else
+				ply:ChatPrint("[red]You must add a minimum of 1000 points to the bounty")
+			end
+		elseif bounty > 0 then
+			BotSayGlobal(to:Nick().."'s bounty is [rainbow]"..bounty.." [fbc]points")
+		elseif bounty == 0 then
+			ply:ChatPrint("[fbc]"..to:Nick().." has no bounty")
+		end
+	elseif
```
swampservers: elseif count>1

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
...
+		ply:ChatPrint("[red]You don't have enough points")
+	end
+end
+
+RegisterChatCommand({'bounty'},function(ply,arg)
+	local t = string.Explode(" ",arg)
+	local to = PlyCount(t[1])
+	if type(to[1]) == "Player" and not to[2]>1 then
+		local bounty = GetPlayerBounty(to)
+		local p = tonumber(t[2])
```
swampservers: An issue here is, what if someone's username is, for example, "trump 6000000", and someone types !bounty trump 6000000 1000 intending to put a 1000 point bounty. A decent fix here would be to use t[#t] instead. Although I think you should actually make two commands, !bounty to see what the bounty is and !setbounty to actually charge points

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
...
+		if type(p) == "number" then
```
swampservers: if p!=nil (it will be a number or nil)

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
...
+		BotSayGlobal("[fbc]"..target:Nick().."'s bounty is now [rainbow]"..GetPlayerBounty(target) + amount.." [fbc]points")
+	else
+		ply:ChatPrint("[red]You don't have enough points")
+	end
+end
+
+RegisterChatCommand({'bounty'},function(ply,arg)
+	local t = string.Explode(" ",arg)
+	local to = PlyCount(t[1])
+	if type(to[1]) == "Player" and not to[2]>1 then
```
swampservers: just do: if count==1

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
...
+		local bounty = GetPlayerBounty(to)
+		local p = tonumber(t[2])
+		if type(p) == "number" then
+			if p > 1000 then
+				AddBounty(ply,to,p)
+			else
+				ply:ChatPrint("[red]You must add a minimum of 1000 points to the bounty")
+			end
+		elseif bounty > 0 then
+			BotSayGlobal(to:Nick().."'s bounty is [rainbow]"..bounty.." [fbc]points")
```
swampservers: did you mean for this to be global? if so i guess thats fine because of the throttle thing

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
...
+	local t = {}
+	for k,v in pairs(player.GetHumans()) do
+		local bounty = GetPlayerBounty(v)
+		if bounty > 0 then
+			table.insert(t,{v,bounty})
+		end
+	end
+	table.sort(t,function(a,b) return a[2] > b[2] end)
+	ply:ChatPrint("[fbc]--- [gold]Bounties [fbc]---")
+	for k,v in pairs(t) do
```
swampservers: do something so it only prints the top 10 or whatever

lua/weapons/weapon_golfclub/shared.lua
```diff
@@ -211,268 +208,269 @@ function SWEP:SetupDataTables()
...
+								self:GetOwner():PrintMessage( HUD_PRINTTALK, "[red]Congratulations! You won [gold]"..tostring(v[2]).." points!" )
+								self:GetOwner():PS_GivePoints(v[2])
+							end
+							iwonsomething = true
+						end
+					end
+				end
+			else
+				if monthlyrecord==1 then strokes = " stroke" end
+				self:GetOwner():PrintMessage( HUD_PRINTTALK, "[red]Your all-time record for this hole is: [orange]"..tostring(record)..strokes )
```
swampservers: Do this with one line please:
"Your all-time record for this hole is X (X this month)"

swampservers (changes requested):
few comments, its almost there

lua/autorun/server/bounties.lua
Noztik: I think sticking to just !bounty is better because it's more intuitive and isn't this how it is for !givepoints?

swampservers: It's not the same because !givepoints only has one function. This one both checks the current bounty AND adds points to the bounty. I think it should be !setbounty so it is explicit that you're trying to set the points (from experience i am quite sure that with this someone would accidentally set a huge bounty sooner or later and demand their points back...)

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,74 @@
...
```
swampservers: Use ipairs to iterate in order, pairs might not be in order

lua/autorun/server/bounties.lua
swampservers: btw, if you just do amount = table.remove(t) and then string.Implode t with a space, you'll get the whole argument back except for the last number. That way if someone's name is "T R U M P" it won't just search for T and fail

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,72 @@
...
+	end
+	return ply.bounty
+end
+
+function SetPlayerBounty(ply,bounty)
+	ply:SetPData("bounty",bounty)
+	ply.bounty = bounty
+end
+
+function AddBounty(ply,target,amount)
```
swampservers, line 26: Can you make it so every player can only set max 10mil of bounties per server restart? Just make a table of steamid->amount set so far. Also decrement the table when they claim a bounty.

lua/autorun/server/bounties.lua
Noztik: how about something like !showbounty, I'd like to keep the addition to bounties tied to !bounty

lua/autorun/server/bounties.lua
Noztik: I don't know how I didn't know that lol

lua/autorun/server/bounties.lua
swampservers: Sure

lua/autorun/server/bounties.lua
```diff
@@ -23,39 +27,57 @@ end
 function AddBounty(ply,target,amount)
 	if !ply:IsPlayer() or !target:IsPlayer() or type(amount) != "number" then return end
 	amount = amount > 0 and amount or 0
-	if ply:PS_HasPoints(amount) then
-		SetPlayerBounty(target,GetPlayerBounty(target) + amount)
+	if BountyLimit[ply:SteamID()] == nil then BountyLimit[ply:SteamID()] = 0 end
+	local total = BountyLimit[ply:SteamID()] + amount
```
swampservers: you can just do like (BountyLimit[ply:SteamID()] or 0) to have a default value

lua/autorun/server/bounties.lua
```diff
@@ -23,39 +27,57 @@ end
...
+	local to,c = PlyCount(arg)
+	if to:IsPlayer() and c == 1 then
+		local bounty = GetPlayerBounty(to)
+		if bounty > 0 then
+			ply:ChatPrint("[orange]"..to:Nick().."'s bounty is [edgy]"..bounty.." [orange]points")
+		else
+			ply:ChatPrint("[orange]"..to:Nick().." has no bounty")
+		end
+	elseif c>1 then
+		ply:ChatPrint("[red]More than one person found with that string in their name")
```
swampservers: you probably don't need to check this here since it's just informational

lua/autorun/server/bounties.lua
```diff
@@ -1,8 +1,12 @@
 
+BountyLimit = {}
+
 hook.Add("PlayerDeath","BountyDeath",function(ply,infl,atk)
 	local bounty = GetPlayerBounty(ply)
+	if BountyLimit[ply:SteamID()] == nil then BountyLimit[ply:SteamID()] = 0 end
 	if bounty > 0 and ply != atk and atk:IsPlayer() then
 		SetPlayerBounty(ply,0)
+		BountyLimit[ply:SteamID()] = math.Clamp(BountyLimit[ply:SteamID()] - bounty,0,10000000)
```
swampservers: can use math.max instead of math.clamp because this only lowers the bounty limit

lua/autorun/server/bounties.lua
```diff
@@ -1,8 +1,12 @@
 
+BountyLimit = {}
```
swampservers: Generally do this like:
BountyLimit = BountyLimit or {}

That way if the file is reloaded it doesn't reset the BountyLimit

lua/autorun/server/bounties.lua
```diff
@@ -23,39 +27,57 @@ end
...
+		ply:ChatPrint("[red]You have reached your limit for today")
 	else
 		ply:ChatPrint("[red]You don't have enough points")
 	end
 end
 
 RegisterChatCommand({'bounty','setbounty'},function(ply,arg)
 	local t = string.Explode(" ",arg)
-	local to,c = PlyCount(t[1])
+	local to,c = PlyCount(string.Implode(table.remove(t)))
```
swampservers: doesnt work. table.remove(t) returns the item which was removed and modifies the table in place. so use whatever table.remove returns as the amount of points.

swampservers (changes requested):
functionality problem

lua/autorun/server/bounties.lua
Noztik: I don't understand what you want me to do with it

Noztik: I don't know why I would remove that check since I only want that message appearing under that circumstance and I have an else for a different message

swampservers: I mean, if there are multiple  people which match you can just post any one of them. the plycount thing is intended to ensure you don't accidentally send points to the wrong person but it's not necessary to be as certain when you're just printing information

lua/autorun/server/bounties.lua
```diff
@@ -44,10 +44,10 @@ end
 
 RegisterChatCommand({'bounty','setbounty'},function(ply,arg)
 	local t = string.Explode(" ",arg)
-	local to,c = PlyCount(string.Implode(table.remove(t)))
+	local p = tonumber(table.remove(t))
+	local to,c = PlyCount(string.Implode(t))
```
swampservers: to will be nil if there is not a matching player. you can't call :IsPlayer on nil. Just change the if statement to if c==1 and assume to is a player

lua/autorun/server/bounties.lua
```diff
@@ -27,7 +27,7 @@ end
 function AddBounty(ply,target,amount)
 	if !ply:IsPlayer() or !target:IsPlayer() or type(amount) != "number" then return end
 	amount = amount > 0 and amount or 0
-	if BountyLimit[ply:SteamID()] == nil then BountyLimit[ply:SteamID()] = 0 end
+	BountyLimit[ply:SteamID()] = BountyLimit[ply:SteamID()] or 0
```
swampservers: i mean you can remove this line (and the one in playerdeath) and replace all non-assignment references to BountyLimit[ply:SteamID()] with (BountyLimit[ply:SteamID()] or 0)

lua/autorun/server/bounties.lua
```diff
@@ -0,0 +1,96 @@
...
+	amount = amount > 0 and amount or 0
+	BountyLimit[ply:SteamID()] = BountyLimit[ply:SteamID()] or 0
+	local total = BountyLimit[ply:SteamID()] + amount
+	if ply:PS_HasPoints(amount) and total <= 10000000 then
+		local add = GetPlayerBounty(target) + amount
+		BountyLimit[ply:SteamID()] = BountyLimit[ply:SteamID()] + amount
+		SetPlayerBounty(target,add)
+		ply:PS_TakePoints(amount)
+		BotSayGlobal("[fbc]"..target:Nick().."'s bounty is now [rainbow]"..add.." [fbc]points")
+	elseif ply:PS_HasPoints(amount) and total > 10000000 then
```
swampservers: the "total > 10000000" check is redundant

swampservers (changes requested):
:IsPlayer functionality problem

lua/autorun/server/bounties.lua
```diff
@@ -46,7 +44,7 @@ RegisterChatCommand({'bounty','setbounty'},function(ply,arg)
 	local t = string.Explode(" ",arg)
 	local p = tonumber(table.remove(t))
 	local to,c = PlyCount(string.Implode(t))
```
swampservers: the seperator should be " ", the default is ""

lua/autorun/server/bounties.lua
```diff
@@ -46,7 +44,7 @@ RegisterChatCommand({'bounty','setbounty'},function(ply,arg)
 	local t = string.Explode(" ",arg)
 	local p = tonumber(table.remove(t))
```
swampservers, line 45: if arg is empty, table.remove here would be an error. I think if arg is empty (#t==0) or if p is nil you should print the usage, and where you currently print the usage you should print "Player (playername from string implode) not found"

lua/autorun/server/bounties.lua
Noztik: my bad, I forgot it needed the argument

swampservers:
Great work, let's try it. I'll send $100 once we know it works.

PR #40, opened 2019-07-12
removed shitty performance impacting meme

Milwacko:


swampservers:
This would just make it an error, you need to remove the spawn code which I think is here

PR #46, opened 2019-07-20
Serverside Error Fixes

chev2:
This should fix some serverside errors

### DIFF
--- a/lua/autorun/server/owoifier.lua
+++ b/lua/autorun/server/owoifier.lua
@@ -6,12 +6,12 @@ local function OwOifier(text)
 	local splitstring = string.Split(text, "")
 	local isemoteenabled = false
 
-	if string.find("!/", splitstring[1]) then
+	if string.find(splitstring[1], "[!/]") then
 		isemoteenabled = true
 	end
 	
 	for k, v in pairs(splitstring) do
-		if string.find(":;[]", v) then --Keep emotes and chat colors the same
+		if string.find(v, "[:;%[%]]") then --Keep emotes and chat colors the same
 			isemoteenabled = !isemoteenabled
 		elseif v == " " then
 			isemoteenabled = false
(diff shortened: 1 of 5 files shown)


lua/weapons/weapon_autism/shared.lua
```diff
@@ -19,7 +19,11 @@ function SWEP:PrimaryAttack()
 		self:ExtEmitSound(soundfile, {speech=-1, pitch=pit})
 		local stime = SoundDuration(soundfile)*100/pit
 		self.Owner:SendLua("util.ScreenShake(LocalPlayer():GetPos(), 5, 0.5, "..tostring((stime*1.65))..", 10000 )")
-		timer.Simple(math.Rand(0,1), function() setPlayerGesture(self.Owner, GESTURE_SLOT_ATTACK_AND_RELOAD, table.Random(autisticTwitches), true ) end)
+		if IsValid(self) and IsValid(self.Owner) then
+			timer.Simple(math.Rand(0,1), function()
```
swampservers: This doesn't fix it, you need to check isvalid inside the timer. The issue happens because the player dies between the timer creation and the function call

lua/autorun/server/owoifier.lua
```diff
@@ -6,12 +6,12 @@ local function OwOifier(text)
 	local splitstring = string.Split(text, "")
 	local isemoteenabled = false
 
-	if string.find("!/", splitstring[1]) then
```
swampservers, line 9: Why change this?

chev2: The arguments were backwards. The first one is a "haystack" the second one is the "needle". All the errors were due to the input string using patterns when they're not supposed to. It still works, I tested it.

swampservers: Oh I see. Would need to do nopatterns=true to have the charset as the haystack

swampservers:
Thank you!

PR #62, opened 2019-08-04
Doomguy permanent (colorable) model

chev2:
The model itself is 9k tris. If you want me to lower it, that's fine.
The actual Doomguy material itself is about 1.3mb, on a 1024x1024 dxt5 image. I can lower it down to 512x512, but that low of a resolution makes the model look like SHIT. if it's absolutely required to increase performance, though, I can lower it.


swampservers:
The idea about perma vs one time models is that perma models should be restricted to "types" of character like pony, ogre etc while one-times can be specific characters like garfield. In this case I think it would be better for this to be a one-time model so there aren't 50 doomslayers running around.

chev2:
That makes sense. I'll keep it in mind. I'm going to leave the colorable part of the model enabled even as a one-time model if that's alright

PR #63, opened 2019-08-05
the minecraft model with changeable middle texture

Milwacko:
it has no lua, wait on chev for that :]

This model uses a 64x64 ABGR8888 texture, for the skin, and hats layer, and a 1x1 DXT1 black texture under the main layer to prevent the model from being completely invisible if a player tries to upload a blank png.

Mesh itself is 1 armature, 1 object, at 576 tris.


swampservers:
Could you do two material layers where the inside skin is opaque and the outer hat layer is alphatest? In code we can just create two mats from one imgur texture so the user won't see.

swampservers:
@Milaco1122 still waiting for update, please remove the inner black layers and put the other two layers on different sub materials

Milwacko:
Do you want the hat and base skin on different materials, or the same like
it is now?

swampservers:
Different materials. In the code, we will use a translucent material for the hat layer and an opaque one for the base layer. These 2 materials will use the same _texture_ but only the hat layer will allow transparency

PR #64, opened 2019-08-06
Minecraft lua + some updates to milaco's materials

chev2:
There is a lot going on here, as I had to accommodate for a lot of stuff, like players changing models needs to reset their material, resetting their skin needs to be networked to everyone, etc.

The only issue with setting materials as far as I know, is if you change models through pointshop with a custom minecraft skin equipped, it won't change the custom material, because I don't know what hook can be used to check the model when it's bought from the pointshop (Use a timer, maybe?). It does reset their model when they die and respawn, though.

The base.vtf doesn't need to be used since we have tools/toolsblack. I also changed the minecraft_pm.vtf to a steve model


chev2:
I still need to add support for ImgurMaterial() btw

swampservers:
Wait did you write all of this or is it from workshop? All of the goofy material code is abstracted when you use imgur material. It should be quite simple to do with a NWString on the player

chev2:
The workshop addons I found with the skin changer had a bunch of lua code so I wrote my own. The material thingy works fine right now, it applies materials just fine without errors, but I can change it over to using imgurmaterial if it's needed

chev2:
I assume ImgurMaterial returns an IMaterial. Since I'm using SetMaterial, it requires a string; would doing ImgurMaterial(args):GetString("$basetexture") return the path to the material?

swampservers:
Basetexture would be the texture, but imaterial should have a GetName or something. For rendering players though try render.MaterialOverride in PrePlayerDraw

chev2:
Also, here's what Milaco's model looks like when you apply a fully transparent texture:

There is a 3rd layer underneath the opaque and hat layer which cannot have its material changed

swampservers:
There shouldn't be three layers. There should two layers of mesh, with a different submaterial on each, and then in code we can ensure that the texture on the lower layer is opaque.

lua/autorun/client/cl_mcsp.lua
```diff
@@ -0,0 +1,98 @@
+local MCSPmodel = "models/milaco/minecraft_pm/minecraft_pm.mdl"
+CreateClientConVar("mcsp_url", "", true, true)
+
+local function MCSPRenderNewMaterial(url, ply) --This is where the magic happens, baby
+	ply.MCSPSkinMaterial = ImgurMaterial(url, ply, ply:GetPos(), false, "VertexLitGeneric", {
```
swampservers: You dont' need to cache this. Just call ImgurMaterial right in PrePlayerDraw. It expects you to keep calling it and it will return different materials as the skin loads. It needs you to keep calling it so it knows when the material is no longer in use as well.

swampservers (changes requested):
unnecessary networking code and won't work because chatcommands are on the server only

swampservers:
What the fuck, github just lost all my comments

lua/autorun/client/cl_mcsp.lua
```diff
@@ -0,0 +1,98 @@
...
+		["$alphatest"] = 1,
+		["$halflambert"] = 1,
+		["$nodecal"] = 1,
+		["$model"] = 1
+	})
+end
+
+RegisterChatCommand({'mcsp','skinpicker'}, function(ply, arg)
```
swampservers: Chatcommands are serverside only.

What you probably want is a command like "!minecraftskin (url)"

What does "MCSP" mean? Using an internal name for your code like "sps" is fine but no player is gonna know it.

lua/autorun/client/cl_mcsp.lua
```diff
@@ -0,0 +1,98 @@
...
+
+			render.MaterialOverrideByIndex(0, v.MCSPSkinMaterial)
+			ragent:DrawModel()
+			ragent:SetNoDraw(true)
+			render.MaterialOverrideByIndex(0, "")
+		end
+	end
+end)
+
+net.Receive("MCSPToCLUpdateNewMaterial", function() --Received from server after a player sends their skin URL query to the server
```
swampservers: You don't need any of this networking junk. Just set the NWString on the player from serverside, and to ImgurMaterial(ply:GetNWString(... from the render hook on the client. It will automatically update and work regardless of when the client joins.

lua/autorun/server/sv_mcsp.lua
```diff
@@ -0,0 +1,59 @@
...
+
+net.Receive("MCSPToSVBroadcastNewMaterial", function()
+	local mcspurl = net.ReadString()
+	local mcspent = net.ReadEntity()
+
+	if (mcspent.SetMCSPTimeout or 0) > CurTime() - 10 then mcspent:Notify("Cooldown...") return end --cooldown
+	mcspent.SetMCSPTimeout = CurTime()
+
+	if mcspent:GetModel() != MCSPmodel then return end
+	if string.len(mcspurl) > 36 then netply:ChatPrint("URL is too long!") return end
```
swampservers: Use the function SanitizeImgurId() here. Takes a string. Returns just the "abcd.png" part of the url or nil if it is bad. The ImgurMaterial thing on clientside expects to get just the "abcd.png" part so they go together

lua/autorun/client/cl_mcsp.lua
swampservers: The function also internally caches and reuses all the data so like, it's fast enough to call every frame with no issue at all. I really should move it into the public code since its clientside anyway

lua/autorun/client/cl_mcsp.lua
chev2: it stands for Minecraft Skin Picker, because I was too lazy to type !minecraftskinpicker into chat when testing

chev2:
You're going to have to talk to Milaco about the model, since he's the one who made it.

At the moment, there is no way to store a player's URL when exiting/joining the server so that it auto-applies on connect.

swampservers:
you can just use PData for it for now

swampservers:
milaco needs to update his thing, we can't render three playermodels as a shitty hack to prevent translucent textures when we can just make the material opaque anyway

lua/autorun/client/cl_mcsp.lua
```diff
@@ -0,0 +1,36 @@
...
+			local ragent = v:GetRagdollEntity()
+
+			render.MaterialOverrideByIndex(0, ImgurMaterial(ply:GetNWString("MCSPSkinURL", false), ply, ply:GetPos(), false, "VertexLitGeneric", {
+				["$alphatest"] = 1,
+				["$halflambert"] = 1,
+				["$nodecal"] = 1,
+				["$model"] = 1
+			}))
+			ragent:DrawModel()
+			ragent:SetNoDraw(true)
```
swampservers: can you please test that the ragdolls still work if you change your model after you respawn? ie it doesnt use the same entity which is still NoDraw'd next time you die

lua/autorun/server/sv_mcsp.lua
```diff
@@ -0,0 +1,35 @@
...
+
+	if ply:GetModel() != MCSPmodel then return end
+	local url = SanitizeImgurId(url)
+	if !url then return ply:ChatPrint("[red]Invalid URL!") end
+
+	ply:SetNWString("MCSPSkinURL", url)
+	ply:SetPData("MinecraftURL", url)
+end
+
+RegisterChatCommand({'minecraftskin','skinpicker','minecrafturl'}, function(ply, arg)
```
swampservers, line 15: The trouble here is people still aren't going to know about this command's existence. Do you think you can make it so when you click "edit playermodel" in the pointshop it pops up a derma string request for the skin url? It should be pretty easy to do. It's possible that i haven't put the clientside pointshop code in here yet but i think i did

chev2: Yes, I assume it would be done in a similar manner to how "customize playermodel" changes to "customize pony" when you've got the pony model equipped, correct?

lua/autorun/client/cl_mcsp.lua
chev2: I tried my best to recreate ImgurMaterial and tested the skins on a bunch of bots constantly changing skin URLs. They kept the URLs even after death and ragdolling a second time, so I'm assuming it works. I copied this bit of code over from one of the minecraft skin pickers on the workshop.

lua/autorun/server/sv_mcsp.lua
swampservers: Yeah. I need to transfer those files over so you can edit them it looks like.

lua/autorun/client/cl_mcsp.lua
swampservers: okay thanks, eventually i will get another repository up for common code like ImgurMaterial

swampservers:
@chev2 I just added the pointshop files in lua/sps. Can you please try to make it so clicking "edit playermodel" pops up a derma string request for the URL?

chev2:
Ok, I updated the lua and materials to match milaco's new model

chev2:
There seems to be an issue with transparency. With the current settings, it seems like the hat layer's transparency renders over the opaque layer of the base layer. I've tried swapping $alphatest and $translucent but they don't seem to fix the issue

lua/autorun/client/cl_mcsp.lua
```diff
@@ -0,0 +1,46 @@
...
+
+hook.Add("PostPlayerDraw", "MCSPRenderSkin", function(ply)
+	if IsValid(ply) and ply:GetModel() == MCSPmodel and ply:IsPlayer() then
+		render.MaterialOverrideByIndex(0, "")
+		render.MaterialOverrideByIndex(1, "")
+	end
+end)
+
+hook.Add("PostDrawOpaqueRenderables", "DrawMinecraftPlayerRagdolls", function() --Renders the material on the player's ragdoll
+	for _, v in pairs(player.GetHumans()) do
```
swampservers: This is too expensive. Can you do Entity:SetMaterial from within https://wiki.garrysmod.com/page/GM/CreateClientsideRagdoll?

materials/models/milaco/minecraft_pm/hat_layer.vmt
```diff
@@ -1,5 +1,7 @@
 "vertexlitgeneric"
 {
     "$basetexture" "models\milaco\minecraft_pm\hat_layer"
-    "$smooth" "0"
```
swampservers, line 4: are these changes still up to date?

lua/autorun/server/sv_mcsp.lua
```diff
@@ -0,0 +1,35 @@
...
+	end
+end, {global=false, throttle=true})
+
+RegisterChatCommand({'mcreset','minecraftreset','minecraftskinreset'}, function(ply, arg)
+	ply:SetNWString("MCSPSkinURL", false)
+	ply:RemovePData("MinecraftURL")
+end, {global=false, throttle=true})
+
+hook.Add("PlayerInitialSpawn", "MinecraftFindURL", function(ply)
+	if ply:GetModel() == MCSPmodel and ply:GetPData("MinecraftURL", false) then MinecraftUpdateURL(ply, ply:GetPData("MinecraftURL", false)) end
```
swampservers: this will break if they join with a different model then change to the mc model. Probably just do ply:SetNWString("MCSPSkinURL", ply:GetPData("MinecraftURL")) for every player

swampservers:
almost there now thanks

materials/models/milaco/minecraft_pm/hat_layer.vmt
chev2: Yes, I made the changes after the new minecraft model PR was accepted

lua/autorun/client/cl_mcsp.lua
```diff
@@ -0,0 +1,38 @@
+local MCSPmodel = "models/milaco/minecraft_pm/minecraft_pm.mdl"
+
+hook.Add("PrePlayerDraw", "MCSPRenderSkin", function(ply) --Material ID 0 is the hat layer, ID 1 is the base layer
+	if IsValid(ply) and ply:GetModel() == MCSPmodel and ply:IsPlayer() then
+		render.MaterialOverrideByIndex(1, ImgurMaterial(ply:GetNWString("MCSPSkinURL", false), ply, ply:GetPos(), false, "VertexLitGeneric", {
```
swampservers, line 5: i'm not sure what will happen if the NWString("MCSPSkinURL") is unset and returns false, can you check for that in the first line of this function and not override if so?

lua/autorun/client/cl_mcsp.lua
```diff
@@ -0,0 +1,38 @@
...
+			["$alpha"] = 1,
```
swampservers: dont use alpha here, if this is the base layer

lua/autorun/client/cl_mcsp.lua
```diff
@@ -0,0 +1,38 @@
...
+			["$halflambert"] = 1,
+			["$model"] = 1
+		}))
+		/*render.MaterialOverrideByIndex(0, ImgurMaterial(ply:GetNWString("MCSPSkinURL", false), ply, ply:GetPos(), false, "VertexLitGeneric", {
+			["$alphatest"] = 1,
+			["$halflambert"] = 1,
+			["$model"] = 1
+		}))*/ --Leave the hat layer out until we figure out what is causing the render issues
```
swampservers: what are the render issues? if you used to use $alpha for both layers there will be render issues where sometimes the base draws over the hat because $alpha doesn't z-write. it should be fine using $alphatest. either way I would prefer to have the hat layer in before releasing this

chev2: There seems to be some sort of issue when setting the base layer to be fully opaque and the hat layer to use $alphatest. If I _only_ render the base layer, it works fine. If I render both the base and hat layer, it's almost like the transparent bits of the hat takes priority over the transparent bits of the base layer that were changed to be fully opaque. It's kind of hard to explain.


The first image is with just the base layer, and the second is both the base layer and hat layer.
I really can't figure out why it's causing this, I've tried changing render order, material parameters, etc. and nothing has fixed it thus far.

chev2: I guess it should be noted that on this skin in particular, there is NOTHING on the hat layer. It's being used as a test to make sure the base layer stays opaque.

swampservers: are you sure it has the issue when using $alphatest (NOT $translucent)?

chev2: Yes. I've always had the hat layer set to $alphatest. I've tried setting the base layer to $opaque 1 or $translucent 0 or whatever disables transparency on the material. I have tried using $translucent on the hat layer before just to see what would happen, and it did cause things to sort of render 'backwards' like you were talking about.

swampservers: What do you mean changing render order? The game should call PrePlayerDraw on its own. 

What's the render group (ENT:GetRenderGroup) of the entity?

chev2: I meant changing render order by switching which render.MaterialOverride is called first. Right now it should be the base layer (ID 1) then the hat layer (ID 0).

The render group for the player is ID 7 (RENDERGROUP_OPAQUE).

swampservers: render.MaterialOverride order doesn't matter, all that does is set a value in a registry.

Can you try adjusting both the .vmt default materials (they might affect model properties or something) as well as the imgurmaterial materials:

- for the base layer don't say anything about alphatest or translucent, the default is opaque
- for the hat layer try translucent instead of alpha test

chev2: Weirdly enough, changing the default materials makes the base layer fully render opaque, and the entire skin is functional. No weird backwards rendering.
However, that's only for the default materials. Trying to apply a custom skin still has the issue, even if I change the hat layer to use $translucent and not specifying any sort of $alphatest or $translucent on the base layer.

swampservers: so if you setup the default materials like that and then apply a custom material with the same params at the same time, it doesn't work? i might have to say just commit it like that and ill check it out on the server later

chev2:
I made the hat_layer.vmt point to minecraft_pm.vtf since the materials are based off the same skin anyway. Opaque on the base layer/alphatest on the hat layer will still work

lua/autorun/client/cl_mcsp.lua
swampservers: Okay thanks, are you ready to merge?

swampservers:
Okay thanks, are you ready to merge?

chev2:
Yes, everything looks good to me. Hopefully we can figure out the hat layer issue ASAP.

PR #65, opened 2019-08-15
Chat Mute

Noztik:
reminder that you need to actually create a network string in the line 220 function for it to work


gamemodes/cinema/gamemode/modules/scoreboard/cl_playerlist.lua
```diff
@@ -203,6 +216,20 @@ function PLAYER:UpdatePlayer()
 			self.Mute:SetImage( "icon32/unmuted.png" )
 		end
     end
+	
+	self.ChatMute.DoClick = function() 
+		self.Player.IsChatMuted = true
```
swampservers: Please allow unmuting

gamemodes/cinema/gamemode/modules/scoreboard/cl_playerlist.lua
```diff
@@ -203,6 +216,20 @@ function PLAYER:UpdatePlayer()
...
+		print("muted "..self.Player:Nick())
+		net.Start("SetChatMuted")
+		net.WriteEntity(self.Player)
+		net.WriteBool(self.Player.IsChatMuted)
+		net.SendToServer()
```
swampservers: Just remove this net it will be purely clientside

gamemodes/cinema/gamemode/modules/scoreboard/cl_playerlist.lua
```diff
@@ -203,6 +216,16 @@ function PLAYER:UpdatePlayer()
 			self.Mute:SetImage( "icon32/unmuted.png" )
 		end
     end
+	
+	self.ChatMute.DoClick = function() 
+		self.Player.IsChatMuted = !self.Player.IsChatMuted or true
```
swampservers: still doesn't work. if IsChatMuted is true, then !IsChatMuted is false, and the "or" will choose true. Just do IsChatMuted=not IsChatMuted

Noztik: but IsChatMuted would be null initially

swampservers: the not operator has higher precedence than the or operator, so its (not IsChatMuted) or true which will always equal true

gamemodes/cinema/gamemode/modules/scoreboard/cl_playerlist.lua
```diff
@@ -203,6 +216,20 @@ function PLAYER:UpdatePlayer()
 			self.Mute:SetImage( "icon32/unmuted.png" )
 		end
     end
+	
+	self.ChatMute.DoClick = function()
+		if !IsValid(self.Player.IsChatMuted) then
```
swampservers: still wrong: https://wiki.garrysmod.com/page/Global/IsValid

IsValid is only for entities, this is just a boolean/nil which i'm pretty sure will never pass the IsValid check (NULL Entity and nil are not the same thing). Just do IsChatMuted = not IsChatMuted

PR #66, opened 2019-08-16
Krusty Hat 

Milwacko:
Krusty Hat sitting at 712 Tris, using 512x512 DXT1 Animated Textures.


swampservers:
this is the greatest pull request of All Time

PR #71, opened 2019-08-23
New minecraft model

Milwacko:
Opaque black layer under base skin removed, and hat layer on seperate material, but same UV


swampservers:
Thank you

PR #74, opened 2019-08-24
Golf clubs can be picked up from the racks

chev2:
People have been asking this for a while. Instead of having the golf club auto-equip, make it so that a player has to actually walk up to the rack to grab one. I added some more golf racks onto the map (7 in total now) so that there are some in the caverns as well.

If you can please remove whatever lua code causes the auto-equip, please do.

I also want to make this as newbie-friendly as possible. A chat message when walking into one of the golf areas or a ply:Notify saying to grab a golf club from one of the racks should work.

### DIFF
--- a/lua/entities/golfrack.lua
+++ b/lua/entities/golfrack.lua
@@ -1,0 +1,84 @@
+if SERVER then AddCSLuaFile() end
+ENT.Type = "anim"
+DEFINE_BASECLASS("base_gmodentity")
+ENT.Model = Model("models/swamponions/golfstand.mdl")
+
+function ENT:Initialize()
+    self:SetModel(self.Model)
+    self:SetMoveType(MOVETYPE_NONE)
+	self:PhysicsInitStatic(SOLID_VPHYSICS)
+
+    self:DrawShadow(false)
+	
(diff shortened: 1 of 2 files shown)


lua/entities/golfrack.lua
```diff
@@ -0,0 +1,84 @@
...
+	end
+end
+
+function ENT:Draw()
+    self:DrawModel()
+end
+
+function ENT:Think()
+    if SERVER then
+        for k, v in pairs(ents.FindInSphere(self:GetPos(), 60)) do
```
swampservers: This is way too expensive. Can you do it with entity:Touch or :Use instead?

swampservers:
Great change thanks!

lua/entities/golfrack.lua
chev2: Sure. That's why I put in a NextThink(), I figured it would be performance-heavy

swampservers:
thanks. can you test this out before i remove the code to automatically give you it? you can just /drop the club then try the stands. In particular please test the no-drawed stands that are on the surface to make sure you can still touch them.

PR #76, opened 2019-08-25
!bountyrandom doesn't include protected players

chev2:


lua/chatcommands/bounties.lua
```diff
@@ -83,14 +83,18 @@ RegisterChatCommand({'bountyall','setbountyall'},function(ply,arg)
...
 	else
 		if p >= 1000 then
-			local ranply = table.Random(player.GetHumans())
-			AddBounty(ply,{ranply},p)
+			local ranply = {}
+			for k, v in pairs(player.GetHumans()) do
+				if !Safe(v) then table.insert(ranply, v) end
+			end
+			AddBounty(ply,{table.Random(ranply)},p)
+			table.Empty(ranply)
```
swampservers: this is unneeded - after this function, since ranply is local, it will go out of scope and get deleted after this function. Even if it was global, the table would persist between calls but each call it gets reassigned to an empty table so they old one gets deleted.

PR #78, opened 2019-09-01
Night-time toggle console command

ghost:
ok so i'll apologize for 2 things in advance :

It's my first attempt at glua so it's just a bunch of copypasting and idk if it can work
i don't think if this is getting the file sent to lua/autorun/client like i intended

This is a lua file that is supposed to add a command "cinema_night_toggle"
and it's just supposed to change the sky to nighttime basically, and it being clientside.

I'm 99% sure it doesn't work so feel free to correct it if you think it's worthwile lol


swampservers:
no this code doesnt come anywhere close to working

the main issue with this idea is that you'd want the map lighting to also change when night is toggled, but I don't know if it's possible to have the clients locally select a different lightmap than what the server wants

ghost:
Doesn't the fact that daynight.lua being located in lua/autorun/client and not in lua/autorun/server
implies that the server doesn't care about what lightmap a player wants ?

chev2:
there's probably a server-sided part of this code that isn't on the github. I believe he's talking about [engine.LightStyle](https://wiki.garrysmod.com/page/engine/LightStyle) which for some reason can only be called on the server realm

swampservers:
that file only draws the skybox, it's synced to the LightStyle. I don't think you can change the lightstyle client side

ghost:
oh okay then nevermind
close the pull request if there's no workaround it

swampservers:
yeah i don't think there is

PR #79, opened 2019-09-01
Some golf changes

chev2:
Added some colors to some of the golf messages (e.g. "<player> failed hole") and also fixed the location of one of the golf stands


swampservers:
Thanks!

PR #82, opened 2019-09-17
Darkmode Pointshop

chev2:
A darkmode pointshop with a toggle-able button in the PS menu

The navigation bar is now gray instead of blue-ish on both the light theme and dark theme


lua/sps/cl_init.lua
```diff
@@ -15,14 +15,50 @@ end)
 
 concommand.Add("ps_togglemenu", function(ply, cmd, args) PS_ToggleMenu() end)
 
+PS_DarkMode = true
+
+concommand.Add("ps_toggletheme", function(ply, cmd, args)
+	if PS_DarkMode then
```
swampservers, line 22: Can you factor this out to a function like SetPointshopTheme(bool dark), and then call that function so the defaults are set in the same place so there is no duplicate configuration?

lua/sps/vgui/DPointShopMenu.lua
```diff
@@ -127,8 +129,7 @@ function PANEL:Init()
 	self.navbar = vgui.Create("DPanel", self)
 	self.navbar:SetTall(PS_NAVBARHEIGHT)
 	self.navbar:Dock(TOP)
-	self.navbar:SetBackgroundColor(BrandColorAlternate)
```
swampservers, line 130: Can we keep the colored header please?

swampservers:
Thank you, chev, very cool

lua/sps/cl_init.lua
```diff
@@ -15,14 +15,50 @@ end)
 
 concommand.Add("ps_togglemenu", function(ply, cmd, args) PS_ToggleMenu() end)
 
+PS_DarkMode = true
+
+concommand.Add("ps_toggletheme", function(ply, cmd, args)
```
swampservers: Could this be a convar so it's saved?

lua/sps/cl_init.lua
chev2: Do you mean something like this?
```
function SetPointshopTheme(dark)
    PS_DarkMode = dark  
end  
SetPointshopTheme(true)  
```

chev2: That's without saved convars, it would probably be something like `SetPointshopTheme(convar:GetBool())` once I put that in

chev2: nevermind, I figured it out

lua/sps/cl_init.lua
```diff
@@ -15,14 +15,54 @@ end)
 
 concommand.Add("ps_togglemenu", function(ply, cmd, args) PS_ToggleMenu() end)
 
+CreateClientConVar("ps_darkmode", "1", true)
+function SetPointshopTheme(dark)
+	PS_DarkMode = dark
+	RunConsoleCommand("ps_toggletheme")
+end
+SetPointshopTheme(!GetConVar("ps_darkmode"):GetBool())
```
swampservers: This won't work because you do RunConsoleCommand("ps_toggletheme") before you do concommand.Add("ps_toggletheme"...). It might have worked in testing because the ps_toggletheme command was already defined from a previous execution of the file when you added this function.

What I meant was, use the SetPointshopTheme function to set all the global variables like PS_TileBGColor so you can remove their other definitions (which have the same values copy pasted). Then initially call it with ps_darkmode and add a change callback to that convar to call it again when necessary.

chev2:
Ok, it should be good now

PR #83, opened 2019-09-18
:(

Milwacko:


swampservers:
This one was an ERROR right? Any idea why? It might say something in console about invalid vertices. BTW the commonfiles.txt thing is no longer used

swampservers:
I fixed the uppercase letters issue, let me know if this still needs to be fixed/removed

PR #85, opened 2019-10-05
Swampyhat/UwU Mask/Cat Ears/Top Hat errors fix

chev2:
I changed all of the .mdl files to point to the right model - they were pointing to `models/<model>.mdl` instead of `models/<author>/<model>/<model>.mdl`. I also fixed the krusty hat material since it was broken and also pointing to the wrong file. I also cut the filesize for the tophat materials a tiny amount.


swampservers:
Excellent!

PR #87, opened 2019-10-18
Pumpkins for Halloween

chev2:
Some pumpkin entities that don't really do a whole lot but exist. Just a little bit of extra decoration, something like the christmas presents/tree but it's for halloween instead

There is an os.date check near the top of the file that prevents the entity from being spawned on the map if the month isn't October, so this entity doesn't have to be enabled/disabled manually


swampservers:
Thank you!

PR #88, opened 2019-10-20
Fix Billy Herrington SWEP (again)

chev2:
I updated the pointshop sh_init.lua with something that should allow product.OnBuy to be used on temp playermodels, which should fix the SWEP (and also allow other temp models to use OnBuy, if needed)


swampservers:
Excellent!

chev2:
@swampservers you can merge this

swampservers:
Sorry, I swear I clicked merge

PR #93, opened 2019-11-23
Darkmode Fixes

chev2:
- Fixed the quantity item count still being black when in dark mode
- The customize buttons on items now match the buttons in the bottom panel


swampservers:
Thanks!

PR #94, opened 2019-11-24
Snowball SWEP

chev2:
It's a snowball SWEP that launches snowballs. It auto-equips when walking outside or into the golf area, but only during December. It also pushes players around a little bit when a snowball hits them.


lua/weapons/weapon_snowball.lua
```diff
@@ -0,0 +1,133 @@
...
+
+		if locname == "Outside" or locname == "Golf" and !ply:HasWeapon("weapon_snowball") then
+			ply:Give("weapon_snowball")
+		end
+	end)
+end
+
+if SERVER then --network the player's new color
+	util.AddNetworkString("CLtoSVSnowballColor")
+	util.AddNetworkString("SVtoCLSnowballColor")
```
swampservers: Please use a player:SetNWVector (use CLtoSVSnowballColor to network it from the player with the weapon to the server, then call that function on the server) for the color state. The reason for this is because if a player joins after the color was  set, using a NWVector will handle sending the correct color for them transparently.

lua/weapons/weapon_snowball.lua
```diff
@@ -0,0 +1,133 @@
...
+		local ply = net.ReadEntity()
+		if !IsValid(ply) then return end
+		ply.currentcolor = col
+	end)
+end
+
+function SWEP:Initialize()
+	self:SetWeaponHoldType(self.HoldType)
+	self.Weapon:SetClip1(1)
+	self.Owner.currentcolor = self.Owner.currentcolor or Color(255, 255, 255)
```
swampservers: "currentcolor" is set on the enttable of the player itself, so it should be something like "snowballcolor" to prevent conflicts (but in this case please use NWVector instead, but put "snowball" in the name of the vector for the same reason)

lua/weapons/weapon_snowball.lua
```diff
@@ -0,0 +1,133 @@
...
+SWEP.AutoSwitchTo = false
+SWEP.HoldType = "grenade"
+
+SWEP.Category = "Snowballs"
+SWEP.Spawnable = true
+
+SWEP.Primary.ClipSize = 1
+
+SWEP.Primary.Automatic = true
+SWEP.ThrowSound = Sound("weapons/iceaxe/iceaxe_swing1.wav")
```
swampservers: missing sound?

chev2: It's one of the Half Life 2 sounds (the crowbar swing sound, specifically) so it should be fine

lua/weapons/weapon_snowball.lua
chev2: A NWVector will cause the alpha channel to not do anything, but that should be fine

chev2:
@swampservers

lua/weapons/weapon_snowball.lua
swampservers: Wait why is there an alpha channel? lol

swampservers:
Thank you!

swampservers:
remind me to update the addon later if i forget

lua/weapons/weapon_snowball.lua
chev2: i was thinking of having an alpha channel if players didn't want trails, but the same effect can be achieved by just setting the trail color to black, so it's fine

PR #96, opened 2019-11-26
Soundcloud Fix 2

Noztik:
your ConsoleMessage code straight up didn't work at the IF statement, also your method of getting the title would always return "Soundcloud" as the page's title


swampservers:
Oh I see, I made zero indexing mistake. I don't see what is wrong with the title thing. On the js side that would give document.title which should be what is in that title tag. What are you thinking it does?

I just fixed the zero indexing, i don't want to merge this because the extra http.get you added is extraneous

Noztik:
go to `http://swampservers.net/cinema/soundcloud.html?url=https://soundcloud.com/alfredo-guzman-navarro/cliffs-of-dover-eric-johnson` and look at the console and the title of the page, it's always "Soundcloud"

swampservers:
ah shit sorry I see. Fixed it now.

Noztik:
oh so there was a way to get the title through the soundclound api afterall

PR #105, opened 2019-12-15
Flappy Fedora rainbow trail

chev2:
The way this works was I tied the sprite trail to an invisible entity that constantly re-adjusts its position in a think hook. I also used the THIRDPERSON variable so that it doesn't move around at all when in first person, therefore cannot be seen until you enter third person.


swampservers (changes requested):
Looks good! Did you intend to delete the fedora_rainbowdash_texture?

lua/entities/ent_fedora_point.lua
```diff
@@ -0,0 +1,34 @@
...
+function ENT:Initialize()
+	if CLIENT then
+		self.Entity:SetNoDraw(true)
+	else
+		self.Trail = util.SpriteTrail(self, 0, Color(255, 255, 255), false, 4, 4, 0.2, 1/8, "chev/rainbowdashtrail") --trail is parented to this entity
+	end
+end
+
+function ENT:Think()
+	if SERVER then return end
```
swampservers, line 23: I think you need to set the entity's position to the player's GetPos serverside, because the server doesn't network the existence of new entities that aren't in a client's PVS (ie that the server doesn't think the client can see)

chev2: Ok, fixed. It should network to the player now since it's set to the player's position in the think hook

chev2:
@swampservers I didn't delete the rainbowdash texture, if you check my fork I just decreased the filesize. It's still there.

lua/entities/ent_fedora_point.lua
chev2: nvm it shows up in first person again, i'll have to fix that

chev2: Ok, I made it so the entity networks to clients regardless of if it's in the PVS or not, it's a cheap fix but it should work

swampservers:
Thanks! Let's see if it works and i'll send you the money

PR #111, opened 2020-01-12
Friend icon in scoreboard

chev2:
If a player on the scoreboard is friends with you on steam, it'll show a little icon next to them on the scoreboard.

Example:
playerfriend

Obviously the 'Player' rank isn't real, but it's just to show that it's not in the same spot as the staff tags.


chev2:
Actually I'm going to go ahead and put it right next to their avatar rather than on the right side, it makes more sense that way

Noztik:
what about people with really long names

chev2:
I changed the icon so it's right next to their avatar now.

friendchange

gamemodes/cinema/gamemode/modules/scoreboard/cl_playerlist.lua
```diff
@@ -260,6 +264,10 @@ function PLAYER:SetPlayer( ply )
 	self.Avatar:SetPlayer( ply, 64 )
 	self.Avatar:SetVisible( true )
 
+	if ply:GetFriendStatus() == "friend" then
+		self.FriendIcon:SetImage("chev/icon/friend.png")
```
swampservers: Can you use https://wiki.garrysmod.com/page/Panel/Hide and the opposite function to control the visibility? Having an undefined image panel seems like it might turn into a missing texture or something, it's undefined behavior, and the icon won't dissapear if someone is unfriended.

chev2: Done

PR #112, opened 2020-01-12
HLS support

Noztik:
the webpage code for hls.html

https://pastebin.com/5nsj4nD1

github just completely screws up me trying to have the webpage code in this description so here's a pastebin link to it


swampservers:
Thanks! I'll install the html as is and we can see if it works. Last time there were issues with your HTML, you could check out how i modified soundcloud.html to see how it works now.

PR #115, opened 2020-01-18
Scoreboard partial lag fix

chev2:
I put a cooldown on how often the playerlist has its layout invalidated, so there should be a pretty good FPS improvement. There's other parts of the code that'll probably have to be adjusted too, this is likely the start


chev2:
idk why the older commits are there, just ignore em

swampservers:
Thanks! I think i'll make it update more often than every 3 seconds though. Ideally it'd only update when something changes but there's no clientside hook for a player changing their name so whatever

PR #116, opened 2020-01-20
hls """fix"""

Noztik:
there is no reason I should have to change it to be this way, this is stupid


swampservers:
You could change the HTML page to accept the url parameter, but it's better to use a more consistent interface anyway. I don't think you need to have the timer around QueueJavascript because it should wait for the page to be ready anyway.

PR #117, opened 2020-02-19
Fedora trail fixes

chev2:
Fedora trail is now parented to the player ent instead of the flappy fedora weapon ent; I also added a timer that will remove trails still on players just in case all the other checks failed somehow (checks every 10 seconds)


swampservers:
One issue I see with the backup timer is if a player logs out but their fedora trail is still there, it wouldn't delete it. What if you just make a tick hook in ent_fedora_point that says "if my owner is invalid or doesn't have fedora equipped, delete"

chev2:
Fixed, ENT:Think() on the serverside uses the server tick so I just put it in the think hook

chev2:
@swampservers

lua/entities/ent_fedora_point.lua
```diff
@@ -21,7 +21,13 @@ end
 
 function ENT:Think()
 	local ply = self.Owner
-	if SERVER then return end
+	if SERVER then
+		if !IsValid(self:GetOwner()) then
+			self:Remove()
+		elseif self:GetOwner():GetActiveWeapon():GetClass() != "weapon_flappy" then
```
swampservers: GetActiveWeapon might be invalid

lua/entities/ent_fedora_point.lua
```diff
@@ -21,7 +21,13 @@ end
 
 function ENT:Think()
 	local ply = self.Owner
-	if SERVER then return end
+	if SERVER then
+		if !IsValid(self:GetOwner()) then
```
swampservers: I meant GetActiveWeapon returns nil if the player has no weapon, so it doesn’t work in that case. You can also clean up the if statement like:
“If not isvalid(owner) or not isvalid(owner.weapon) or not owner.weapon.class==fedora then remove”
That works because of short circuit boolean evaluation - if owner is invalid, it won’t run the later statements so it won’t error

swampservers (changes requested):

lua/entities/ent_fedora_point.lua
chev2: yeah that was what I was afraid of, I didn't want to check if the owner was valid and if the owner's weapon was valid on the same statement because it would error if it checked later statements

PR #118, opened 2020-02-29
Coronavirus replaces ebola

chev2:
Might as well, lol

All the variable names and whatnot are still the same so that we don't have to replace all the names to fit


chev2:
if "coronavirus" is too stale, "COVID-19" also works

swampservers:
reddit moment, but alright

PR #120, opened 2020-03-10
Bitchute and Streamwatch

Noztik:
I don't know if there is a better way to get the bitchute mp4 link and carry it down, all I can figure out that 100% works is fetching it from the bitchute page every time I need it

### DIFF
--- a/gamemodes/cinema/gamemode/modules/theater/services/sh_bitchute.lua
+++ b/gamemodes/cinema/gamemode/modules/theater/services/sh_bitchute.lua
@@ -1,0 +1,84 @@
+local SERVICE = {}
+
+SERVICE.Name 		= "Bitchute"
+
+SERVICE.NeedsCodecs = true
+
+function SERVICE:GetKey( url )
+	match = string.match(url.path,"/.+/(.+[^/])")
+	if match != nil and string.find(url.host,"bitchute.com") then
+		return match
+	end
+	return false
+end
+
+if CLIENT then
+	function SERVICE:GetVideoInfoClientside(key, callback)
+		EmbeddedCheckCodecs(function()
+			vpanel = vgui.Create("DHTML")
+		
+			vpanel:SetSize(100,100)
+		
+			vpanel:SetAlpha(0)
+		
+			vpanel:SetMouseInputEnabled(false)
+			
+			timer.Simple(20,function() 
+				if IsValid(vpanel) then
+					vpanel:Remove() 
+					print("Failed")
+					callback()
+				end
+			end)
+		
+			function vpanel:ConsoleMessage(msg)
+				if msg:StartWith("DURATION:") then
+					local duration = math.ceil(tonumber(string.sub(msg,10)))
+					print("Duration: "..duration)
+					self:Remove()
+					print("Success!")
+					callback({duration=duration})
+				end
+			end
+			
+			http.Fetch("https://www.bitchute.com/embed/"..key,
+				function(body,length,headers,code)
+					match = string.match(body,"source src=\"(.+)\" type")
+					if match != nil then
+						vpanel:OpenURL("http://swampservers.net/cinema/filedata.php?file="..match)
+					end
+				end,
+				function(err)
+					print("Failed to reach bitchute page while fetching duration")
+					vpanel:Remove()
+					callback()
+				end
+			)
+		end,
+		function()
+			chat.AddText("You need codecs to request this. Press F2.")
+			return callback()
+		end)
+	end
+	
+	function SERVICE:LoadVideo( Video, panel )
+		http.Fetch("https://www.bitchute.com/embed/"..Video:Key(),
+			function(body,length,headers,code)
+				match = string.match(body,"source src=\"(.+)\" type")
+				if match != nil then
+					local url = "http://swampservers.net/cinema/file.html"
+					panel:OpenURL(url)
+					
+					local k = Video:Key()
+					local str = string.format( "th_video('%s');", string.JavascriptSafe(k) )
+					panel:QueueJavascript( str )
+				end
+			end,
+			function(err)
+				print("Failed to reach bitchute page while loading video")
+			end
+		)
+	end
+end
+
+theater.RegisterService( 'bitchute', SERVICE )
--- a/gamemodes/cinema/gamemode/modules/theater/services/sh_bitchute.lua
+++ b/gamemodes/cinema/gamemode/modules/theater/services/sh_bitchute.lua
@@ -62,22 +62,13 @@ if CLIENT then
 	end
 	
 	function SERVICE:LoadVideo( Video, panel )
-		http.Fetch("https://www.bitchute.com/embed/"..Video:Key(),
-			function(body,length,headers,code)
-				match = string.match(body,"source src=\"(.+)\" type")
-				if match != nil then
-					local url = "http://swampservers.net/cinema/file.html"
-					panel:OpenURL(url)
-					
-					local k = Video:Key()
-					local str = string.format( "th_video('%s');", string.JavascriptSafe(k) )
-					panel:QueueJavascript( str )
-				end
-			end,
-			function(err)
-				print("Failed to reach bitchute page while loading video")
-			end
-		)
+		local url = "http://swampservers.net/cinema/file.html"
+		panel:OpenURL(url)
+		
+		local k = Video:Data()
+		
+		local str = string.format( "th_video('%s');", string.JavascriptSafe(k) )
+		panel:QueueJavascript( str )
 	end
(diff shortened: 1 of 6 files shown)


Noztik:
use `<title>(.*)</title>` to get the bitchute video title btw

gamemodes/cinema/gamemode/modules/theater/services/sh_hls.lua
```diff
@@ -12,7 +12,24 @@ function SERVICE:GetKey( url )
 	if string.sub( url.path, -5) == ".m3u8" then
 		return url.encoded
 	end
-	return false
+	match = string.match(url.encoded,"streamwat.ch/(%w+)/*$")
+	if match != nil and string.find(url.host,"streamwat.ch") then
+		http.Fetch("http://streamwat.ch/"..match.."/player.min.js",
```
swampservers: Won't work. http.Fetch is an asynchronous function, and the "returns" inside its callback will only return from the callback, not return something to the caller. GetKey is meant to be a lightweight function to identify a possibly matching URL anyway and it wouldn't be appropriate to use HTTP here. Let me update the repo with my youtube implementation so you can see how to get data asynchronously

gamemodes/cinema/gamemode/modules/theater/services/sh_bitchute.lua
```diff
@@ -0,0 +1,84 @@
...
+			)
+		end,
+		function()
+			chat.AddText("You need codecs to request this. Press F2.")
+			return callback()
+		end)
+	end
+	
+	function SERVICE:LoadVideo( Video, panel )
+		http.Fetch("https://www.bitchute.com/embed/"..Video:Key(),
```
swampservers: LoadVideo is run clientside for everyone watching. Can this function be pushed back to serverside and run once, or is the resulting link unique to each client?

Noztik: it's not unique to each client, I'm not sure how you'll implement it serverside

gamemodes/cinema/gamemode/modules/theater/services/sh_hls.lua
Noztik: that's why I excluded the return at the end if the link is from streamwat.ch so it will let http.Fetch use the callback correctly

swampservers: I added the serverside of youtube and soundcloud. Youtube is an example of how to use the server to get extra data asychronously; you can use the .data field of the video to store anything arbitrary other than title/duration/thumbnail (which is why I renamed the identifier the "KEY", so I could use "DATA" as the name for the other stuff). The soundcloud serverside is an example of the plumbing needed to use the client for stuff

gamemodes/cinema/gamemode/modules/theater/services/sh_bitchute.lua
swampservers: I added the serverside of youtube and soundcloud. Youtube is an example of how to use the server to get extra data asychronously; you can use the .data field of the video to store anything arbitrary other than title/duration/thumbnail (which is why I renamed the identifier the "KEY", so I could use "DATA" as the name for the other stuff). The soundcloud serverside is an example of the plumbing needed to use the client for stuff

gamemodes/cinema/gamemode/modules/theater/services/sh_hls.lua
swampservers: That's not how it works... if the function "ends" without a "return" statement it implicitly returns nil. The callback part within http.Fetch won't be run until the webpage responds which would be like a second later, and by that time the entire context under which GetKey was called will be gone. If you checkout how youtube's serverside works, the function takes in "OnSuccess" and "OnFailure" arguments, both of which are functions to be run when the website has responded, which will continue the slow asychronous process of requesting the video.

gamemodes/cinema/gamemode/modules/theater/services/sh_bitchute.lua
Noztik: I still need access to the info table in EmbeddedCheckCodecs unless you're fine with using a http function in there, same with the hls EmbeddedCheckCodecs function too

swampservers: Not sure what you mean. If you update the PR to what you think would work, I can let you know if there's a proper way to do it

Noztik: you talked about putting the real url into the .data member/whatever but I can't access that where I actually need it

swampservers: Oh, sorry, so if you look at sv_youtube.lua there's the whole sv_GetVideoInfo.youtube which is the serverside way of getting video data before it's going to play. In that function it populates the "info" table. You can use info.data there to store arbitrary stuff.

Noztik: yes but what good is that when I can't access it anywhere else

Noztik: unless info.key is the same variable being passed to both GetVideoInfoClientside and LoadVideo

swampservers: The Video object should have a Data() field to get it when you call service:LoadVideo. I think a good way to handle this would be to store the actual URL which gets eventually passed to the JWPlayer (ie "file.html") in the Data field instead of using the Key field for it. For this service it's optional, but for the other service you still need to make your asynchronous request inside of a sv_GetVideoInfo function (because GetKey is not asychronous and it won't work the way it is right now)

Noztik: what about the GetVideoInfoClientside function, only the key and callback is passed to it

swampservers: Actually, you're right, if you're doing a GetVideoInfoClientside then with the current plumbing, you can only send the Key to the client. If you were using the Data as the video URL, well, it wouldn't work unless I added support for that.

You can just make HTTP requests in GetVideoInfoClientside like it currently is, its fine. The blocker on this right now is the other comment about the asynchronous return not working

swampservers:
Oh, sv_hls already existed. I just moved it to this addon. sorry about that

gamemodes/cinema/gamemode/modules/theater/services/sh_bitchute.lua
```diff
@@ -0,0 +1,75 @@
...
+function SERVICE:GetKey( url )
+	match = string.match(url.path,"/.+/(.+[^/])")
+	if match != nil and string.find(url.host,"bitchute.com") then
+		return match
+	end
+	return false
+end
+
+if CLIENT then
+	function SERVICE:GetVideoInfoClientside(key, callback)
```
swampservers, line 17: You're not calling theater.GetVideoInfoClientside in the serverside part (see sv_soundcloud), so this entire function is unused

gamemodes/cinema/gamemode/modules/theater/services/sh_bitchute.lua
```diff
@@ -0,0 +1,75 @@
...
+				end
+			)
+		end,
+		function()
+			chat.AddText("You need codecs to request this. Press F2.")
+			return callback()
+		end)
+	end
+	
+	function SERVICE:LoadVideo( Video, panel )
```
swampservers, line 67: I use EnsureURL here, check sh_base.lua (optional)

gamemodes/cinema/gamemode/modules/theater/services/sv_hls.lua
```diff
@@ -1,18 +1,43 @@
...
-
-
-
-			
+
+sv_GetVideoInfo = sv_GetVideoInfo or {}
+
+sv_GetVideoInfo.hls = function(self, key, ply, onSuccess, onFailure)
+
+	local match = string.match(key,"streamwat.ch/(%w+)/*$")
```
swampservers: prefer naming this variable like streamwatch_key or something (its totally not clear in other places that its related to streamwatch)

Noztik:
check the new commit for issues

swampservers:
Do you want to add a bunch of print statements so you can test it? I doubt this is going to work correctly before it's been tested lol. I guess it looks okay to me though

Noztik:
I added clientside print and that ServerDebug function to the serverside files

swampservers:
Thanks!

PR #121, opened 2020-03-13
More themes for the caverns

chev2:
Both 'cavernalt' and 'cavernnight' behave exactly like the current theme, where synchronization doesn't matter, and it loops.

cavernalt - https://youtu.be/-erU20cQO_Y
cavernnight - https://youtu.be/QT8vuiS0cpQ


swampservers:
Thank you!

PR #122, opened 2020-03-13
Area music changes

chev2:
Made it so the caverns/caverns alt. theme during the day won't constantly remove each other when in the caverns


swampservers:
lol nice

PR #125, opened 2020-03-21
Support Update

Noztik:
somehow missed that sh_bitchute had reverted
no clue why this is causing hls to blackscreen


Noztik:
is Video:Data() nil if you don't assign anything to it?

swampservers:
It should be nil yeah. Possibly empty string

PR #126, opened 2020-03-22
hls fix

Noztik:
of course it was a silly mistake like this


Noztik:
so many pull requests for just video support lol

Noztik:
this should be the final thing

PR #131, opened 2020-04-14
hls debug

Noztik:
this is ridiculous, hls is totally broken on the server even though it works perfectly fine in singleplayer and any browser


swampservers:
Does it work on your client's screen when you are the one requesting it? you can ask me on discord and i can help

PR #134, opened 2020-04-17
video debug

Noztik:
optional debug triggered by a console command to view all errors and messages coming from a page on a theater screen
for some reason some things don't work properly only when on screen but work fine in the F1 browser, the request browser, and even on theater screens in private servers


gamemodes/cinema/gamemode/modules/scoreboard/controls/cl_html.lua
```diff
@@ -222,9 +222,16 @@ end
 PANEL.QueueJavaScript = PANEL.QueueJavascript
 PANEL.Call = PANEL.QueueJavascript
 
+concommand.Add("cinema_video_debug",function(ply)
+	ply.videoDebug = not ply.videoDebug
```
swampservers, line 226: this could just be a global VIDEO_DEBUG

PR #143, opened 2020-05-22
Added weapon_tag

Ugleh:
I could not test out ExtEmitSound but everything worked fine when using
EmitSound. Localization might neen to be changed, depending on your
preference. Line 6 for description and Line 90 for overlay.


lua/sps/items/weapons.lua
```diff
@@ -103,6 +103,14 @@ PS_WeaponProduct({
 	class = 'weapon_gauntlet'
 })
 
+PS_WeaponProduct({
```
swampservers: I'm not sure this should be a shop item. I could put it in the mystery box instead (the code for that is private, I can do it)

Ugleh: > 
> 
> I'm not sure this should be a shop item. I could put it in the mystery box instead (the code for that is private, I can do it)

Sounds good, want me to revert the changes or are you? BTW Probably should test the audio levels. It was fine for me but I don't know your standard.

lua/weapons/weapon_tag.lua
```diff
@@ -0,0 +1,150 @@
+
```
swampservers, line 1: Make this weapon un-droppable in https://github.com/swampservers/contrib/blob/master/lua/chatcommands/cinema_simple.lua line 46

lua/weapons/weapon_tag.lua
```diff
@@ -0,0 +1,150 @@
...
+SWEP.DrawAmmo = false
+
+SWEP.HitDistance = 48
+
+
+local justReloaded = 0
+local initializeTime = 0
+local tagTimer = 35
+function SWEP:Initialize()
```
swampservers, line 37: According to the wiki, SWEP:Initialize is sometimes not called clientside, so this is unreliable. Can you instead do self:SetNWFloat("initTime", CurTime()) from the serverside?

lua/weapons/weapon_tag.lua
```diff
@@ -0,0 +1,150 @@
...
+	 if not Safe(target) then
+		 TagPlayer(self,target,attacker)
+	 elseif attacker:GetTheater() and attacker:GetTheater():IsPrivate() and attacker:GetTheater():GetOwner() == attacker and attacker:GetLocationName() == target:GetLocationName() then
+		 TagPlayer(self,target,attacker)
+	 else return end
+end
+
+
+function SWEP:DrawHUD()
+local TextWidth = surface.GetTextSize("Trebuchet24")
```
swampservers: can't you GetTextSize with the message string? Not all characters are the same width

lua/weapons/weapon_tag.lua
```diff
@@ -0,0 +1,150 @@
...
+
+local justReloaded = 0
+local initializeTime = 0
+local tagTimer = 35
+function SWEP:Initialize()
+	self:SetHoldType( "fist" )
+	initializeTime = CurTime()
+	if SERVER then
+		timer.Simple(5,function() if IsValid(self) and self.Owner:IsFrozen() then self.Owner:UnLock() end end)
+		timer.Simple(tagTimer,function() if IsValid(self) then self:Remove() self.Owner:Kill() end end)
```
swampservers: Can you make them explode? (util.Effect, see the jihad/bigbomb code but only make one explosion effect)

lua/weapons/weapon_tag.lua
Ugleh: Your right. I misunderstood the GetTextSize function. Will fix.

lua/weapons/weapon_tag.lua
```diff
@@ -0,0 +1,150 @@
...
+	end
+	justReloaded=2
+end
+
+function SWEP:Tick()
+	justReloaded = justReloaded-1
+end
+
+
+function TagPlayer(self,target,attacker)
```
swampservers: This func name seems like it could cause a conflict, can you make it local please?

lua/weapons/weapon_tag.lua
Ugleh: Will do, appreciate all the comments.

swampservers:
This is really good overall, thank you!

lua/weapons/weapon_tag.lua
Ugleh: I assumed everything encapsulated in the class was already local to the class itself. My bad.

lua/sps/items/weapons.lua
swampservers: Revert in the same PR. I assume the audio is fine, you can revise it later if you want to.

lua/weapons/weapon_tag.lua
swampservers: Nah it's global. You could do SWEP:TagPlayer(target,attacker) though, that would be better anyway

lua/weapons/weapon_tag.lua
Ugleh: This is the only thing I am not wrapping my head around. Because Initialize sometimes doesn't run clientside should I just set initializeTime inside the if SERVER statement? If I use the code you provided is initTime a local variable for that class?

Ugleh:
> 
> 
> This is really good overall, thank you!

I believe it is all done. All the edits you asked for.

lua/weapons/weapon_tag.lua
```diff
@@ -129,16 +138,14 @@ function SWEP:PrimaryAttack( right )
 				end
 			end
 			if (target[2] < 50) then
-				TestTagPlayer(self,target[1],self.Owner)
+				self.TestTagPlayer(self,target[1],self.Owner)
```
swampservers, line 141: You can call a member function with self:TestTagPlayer (colon instead of dot) and it will implicitly pass self as the first argument

Ugleh: That is cool too know. Does a period still work in this case?

swampservers: It works because you passed self in manually, it's just not the "ideal" way to do it

PR #147, opened 2020-05-25
Added coinflip chatcommand game

Ugleh:
1000 point minimum, initiator is always heads, 50/50 coinflip.

!coinflip player points

### DIFF
--- a/lua/chatcommands/coinflip.lua
+++ b/lua/chatcommands/coinflip.lua
@@ -43,8 +43,6 @@ function()
 			elseif(fromPlayer ~= nil and toPlayer == nil) then
 				fromPlayer:ChatPrint("[fbc]The player you requested a coinflip to has left. Try again lates.")
 			end
-			table.remove(CoinFlips, usedID)
-
 		else
 			NewCoinFlips[fromID] = coinflip
 		end
--- a/lua/chatcommands/coinflip.lua
+++ b/lua/chatcommands/coinflip.lua
@@ -1,0 +1,85 @@
+CoinFlips = CoinFlips pr {}
+RegisterChatCommand({'coin','coinflip'},function(ply,arg)
+	if arg:lower() == "accept" then checkCoinFlipRequest(ply) return end
+	local t = string.Explode(" ",arg)
+	local p = tonumber(table.remove(t))
+	if p == nil then
+		ply:ChatPrint("[orange]!coinflip player points")
+	else
+		local to,c = PlyCount(string.Implode(" ",t))
+		if c == 1 then
+			if p >= 1000 then
+				initCointFlip(ply,to,p)
+			else
+				ply:ChatPrint("[red]A coin flip must be a minimum of 1000 points. No pussy bets!")
+			end
+		else
+			ply:ChatPrint("[red]Player "..string.Implode(" ",t).." not found")
+		end
+	end
+end,{global=true,throttle=true})
+
+function initCointFlip(ply,target,amount)
+	if ply:PS_HasPoints(amount) and target:PS_HasPoints(amount) and CoinFlips[ply:SteamID()] == nil then
+		-- Both players have enough points AND an existing coinflip doesn't exist.
+		ply:ChatPrint("[orange]"..to:Nick().." is receiving your coinflip request.")
+		CoinFlips[ply:SteamID()] = {target:SteamID(), amount}
+		timer.Simple(120, CoinFlipRequestTimeout, ply, target)
+	elseif CoinFlips[ply:SteamID()] ~= nil then
+		ply:ChatPrint("[red]What the hell r yeh doing bruv? yeh already hae a filp goin!")
+	elseif not ply:PS_HasPoints(amount) then
+		ply:ChatPrint("[red]What the hell r yeh doing bruv? yeh don't hae dooze points!")
+	elseif not target:PS_HasPoints(amount) then
+		-- Should a player be told another player doesn't have enough points?
+		ply:ChatPrint("[red]y'r bruv doesn't hae dooze points to coinflip with yeh!")
+	end
+end
+
+local function CoinFlipRequestTimeout(ply, target)
+	table.remove(CoinFlips, ply:SteamID())
+	ply:ChatPrint("[fbc]The other bruv don't want to play. Try again lates.")
+	target:ChatPrint("[fbc]Missed out on a coinflip from " .. ply:Nick() .. ".")
+end
+
+function checkCoinFlipRequest(target)
+	coinflipFound = false
+	for i,j in pairs(CoinFlips) do
+		if j[1] == target:SteamID() then
+			-- Coinflip Request Found
+			coinflipFound = true
+			finishCoinFlip(i, target)
+		end
+	end
+	if not coinflipFound then
+		ply:ChatPrint("[red]What the hell r yeh doing bruv? yeh don't hae a coinflip request!")
+	end
+end
+
+function finishCoinFlip(fromID, to)
+	local from = nil
+	local amount = CoinFlips[fromID][2]
+	-- Get Player from ID
+	for _,ply in ipairs( player.GetAll() ) do
+		if(ply:SteamID() == fromID) then from = ply end
+	end
+	if(from == nil) then
+		to:ChatPrint("[red]The other bruv left, gotta cancel this coin flip!")
+	elseif from:PS_HasPoints(amount) and to:PS_HasPoints(amount) then
+		-- Final Check, make sure they have funds still
+		local heads = math.random(0, 1) -- the "request from" player is always Heads.
+		BotSayGlobal("[fbc]"..from:Nick().." flipped a coin worth [rainbow]"..(amount * 2).."[fbc] against "..to:Nick().." and...... " .. ((heads == 1) and "Won" or "Lost") .."!")
+		from:ChatPrint("[fbc]You " .. ((heads == 1) and "Won" or "Lost") .. " " .. amount .. ".")
+		to:ChatPrint("[fbc]You " .. ((heads == 0) and "Won" or "Lost") .. " " .. amount .. ".")
+		-- Instead of taking the amount away from both and then giving the winner the amount x 2, simply remove/add here
+		if heads == 1 then
+			from:PS_GivePoints(amount)
+			to:PS_TakePoints(amount)
+		else
+			from:PS_TakePoints(amount)
+			to:PS_GivePoints(amount)
+		end
+	else
+		from:ChatPrint("[red]One of you no longer have the funds required for this coinflip!");
+		to:ChatPrint("[red]One of you no longer have the funds required for this coinflip!");
+	end
+end
--- a/lua/chatcommands/coinflip.lua
+++ b/lua/chatcommands/coinflip.lua
@@ -10,6 +10,8 @@ RegisterChatCommand({'coin','coinflip'},function(ply,arg)
 		if c == 1 then
 			if ply == to then
 				ply:ChatPrint("[red]You can't coinflip yourself!")
+			elseif hasCoinflipRequest(to) then
+				ply:ChatPrint("[red]That player currently has a coinflip request active!")
 			elseif p >= 1000 then -- minimum 1000 point coinflip
 				initCointFlip(ply,to,p)
 			else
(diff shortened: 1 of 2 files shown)


lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,89 @@
...
+	local amount = CoinFlips[fromID][2]
+	-- Get Player from ID
+	for _,ply in ipairs( player.GetAll() ) do
+		if(ply:SteamID() == fromID) then from = ply end
+	end
+	if(from == nil) then
+		to:ChatPrint("[red]The other bruv left, gotta cancel this coinflip!")
+	elseif from:PS_HasPoints(amount) and to:PS_HasPoints(amount) then
+		-- Final Check, make sure they have funds still
+		local heads = math.random(0, 1) -- the "request from" player is always Heads.
```
swampservers: make heads a bool here: `local heads = math.random() < 0.5`

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,89 @@
...
+		BotSayGlobal("[edgy]" .. from:Nick() .. "[fbc] flipped a coin worth [rainbow]" .. (amount * 2) .. "[fbc] against [gold]".. to:Nick().. "[fbc] and...... [rainbow]" .. ((heads == 1) and "Won" or "Lost") .."!")
+		from:ChatPrint("[fbc]You " .. ((heads == 1) and "Won" or "Lost") .. " " .. amount .. ".")
+		to:ChatPrint("[fbc]You " .. ((heads == 0) and "Won" or "Lost") .. " " .. amount .. ".")
+		-- Instead of taking the amount away from both and then giving the winner the amount x 2, simply remove/add here
+		if heads == 1 then
+			from:PS_GivePoints(amount)
```
swampservers: Take before Give, in case Take errors. If you Give before Take errors, you've got a duping exploit

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,89 @@
...
+		-- Both players have enough points AND an existing coinflip doesn't exist.
+		ply:ChatPrint("[orange]"..to:Nick().." is receiving your coinflip request.")
+		CoinFlips[ply:SteamID()] = {target:SteamID(), amount}
+		timer.Simple(120, CoinFlipRequestTimeout, ply, target) -- 2 minutes for the timeout.
+	elseif CoinFlips[ply:SteamID()] ~= nil then
+		ply:ChatPrint("[red]What the hell r yeh doing bruv? yeh already hae a filp goin!")
+	elseif not ply:PS_HasPoints(amount) then
+		ply:ChatPrint("[red]What the hell r yeh doing bruv? yeh don't hae dooze points!")
+	elseif not target:PS_HasPoints(amount) then
+		-- Should a player be told another player doesn't have enough points?
```
swampservers: No, you can use this to see exactly how much someone has.

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,89 @@
...
+local function CoinFlipRequestTimeout(ply, target)
+	table.remove(CoinFlips, ply:SteamID())
+	ply:ChatPrint("[fbc]The other bruv don't want to play. Try again lates.")
+	target:ChatPrint("[fbc]Missed out on a coinflip from " .. ply:Nick() .. ".")
+end
+
+function checkCoinFlipRequest(target)
+	-- There is probably a better way to do this. I have always found this kind of coding nasty once you get a sip of Java 8.
+	coinflipFound = false
+	for i,j in pairs(CoinFlips) do
```
swampservers: So if multiple people all give one guy requests, it will accept all of them? It should only do the most recently sent one

lua/chatcommands/coinflip.lua
Ugleh: Any recommendation as to resolve this? Should it just act as if the player never accepted and time out accordingly?

lua/chatcommands/coinflip.lua
Ugleh: Should I instead prevent multiple requests to the same person?

lua/chatcommands/coinflip.lua
Ugleh: Should I be calling math.randomseed(os.time()) or is that already called globally somewhere?

Ugleh:
@swampservers Still learning github in regards to using multiple branches for multiple PR's and I don't know what
"Merge remote-tracking branch 'refs/remotes/swampservers/master'" is or if it needs to be reverted to an older commit because of it. It may have been caused when I switched my branches, but that doesn't make sense. Let me know if I need to revert.

swampservers:
I'm pretty sure that just means it put the changes from your other PR into the code state that this PR is coming from

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,102 @@
+CoinFlips = CoinFlips or {}
+RegisterChatCommand({'coin','coinflip'},function(ply,arg)
+	if arg:lower() == "accept" then checkCoinFlipRequest(ply) return end
```
swampservers: Honestly, this design won't work. Say person A makes a request to B to flip for 1000 points.  Then while person B is typing "!coin accept", malicious user C types "!coin B 1000000" and suddenly B loses 1 million. This is 100% something that would happen over and over.

I suggest make the other player type `!coin accept <amount>` to confirm the bet amount

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,102 @@
...
+		end
+	end
+end,{global=true,throttle=true})
+
+function initCointFlip(ply,target,amount)
+	if ply:PS_HasPoints(amount) and target:PS_HasPoints(amount) and CoinFlips[ply:SteamID()] == nil then
+		-- Both players have enough points AND an existing coinflip doesn't exist.
+		ply:ChatPrint("[orange]"..to:Nick().." is receiving your coinflip request.")
+		CoinFlips[ply:SteamID()] = {target:SteamID(), amount}
+		timer.Simple(120, CoinFlipRequestTimeout, ply, target) -- 2 minutes for the timeout.
```
swampservers: This timer never gets removed, so the request timeout thing will screw with later requests. And it will always print that stuff too.

lua/chatcommands/coinflip.lua
Ugleh: I get what you are saying. What about instead of that only 1 request can be made to a person at a time?

lua/chatcommands/coinflip.lua
Ugleh: Totally forgot about that. Thanks for catching that. So the resolution is to make a non-simple timer with a unique identifier, and save the identifier in the CoinFlip?

Ugleh: Something like `timer.Create( "CoinFlip" .. ply:SteamID(), 120, 1, CoinFlipRequestTimeout, ply, target)` and anywhere I do table.remove I will replace with a function that removes the table but also removes the timer?

lua/chatcommands/coinflip.lua
swampservers: then someone can lock it, so no

lua/chatcommands/coinflip.lua
swampservers: A more solid solution would be to use one global timer that runs every second `timer.Create("Coinflips",1,0, ...)` which checks each one in the table. Then have CurTime() written in each table member when you create it, and if the Coinflips timer sees something more than 60 seconds old, delete it

Ugleh: Was thinking about doing this but didn't know if it was the right move or not. You say 60 seconds, is that the timeout you want?

lua/chatcommands/coinflip.lua
Ugleh: Honestly I don't think someone is going to be attempting to lock someone out of a coinflip for 60 seconds. I just think that "!coinflip accept amount" is such a weird command. How about if they only have 1 request they can do "!coinflip accept" but if they have multiple requests they have to do "!coinflip accept amount"

swampservers: Honestly I think you still need to make people type the amount. I don't want people asking me for refunds because "I thought the flip was only for 10k not 100k" and whatnot. It would happen a lot.

lua/chatcommands/coinflip.lua
swampservers: seems fine

Ugleh:
@swampservers I mentioned earlier, but do I need to worry about putting this somewhere? `math.randomseed(os.time())`

swampservers:
Nah, it's global so the state is already unpredictable

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,153 @@
...
+			ply:ChatPrint("[orange]!coinflip accept points")
+		else
+			ply:ChatPrint("[orange]!coinflip player points")
+		end
+	else
+		local to,c = PlyCount(string.Implode(" ",t))
+		if c == 1 then
+			if ply == to then
+				ply:ChatPrint("[red]You can't coinflip yourself!")
+			elseif hasCoinflipRequest(to) then
```
swampservers: Since you're forcing people to confirm the number of points, this is no longer necessary

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,153 @@
+CoinFlips = CoinFlips or {}
+
+RegisterChatCommand({'coin','coinflip'},function(ply,arg)
+	local t = string.Explode(" ",arg)
+	local p = tonumber(table.remove(t))
+	if #t == 1 and t[1]:lower() == "accept" then checkCoinFlipRequest(ply, p) return end
+
+	if p == nil then
+		if #t == 1 and t[1]:lower() == "accept" then
+			ply:ChatPrint("[orange]!coinflip accept points")
```
swampservers: Write: !coinflip accept [confirm number of points]

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,153 @@
...
+			end
+		else
+			ply:ChatPrint("[red]Player "..string.Implode(" ",t).." not found")
+		end
+	end
+end,{global=true,throttle=true})
+
+timer.Create( "CoinFlip", 1, 0,
+function()
+	local usedID = nil
```
swampservers: Make a list of things to remove then iterate that list removing things.

Alternately, make a new empty table and add everything you aren't going to remove to it, then set CoinFlips = (the new table) when you're done.

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,153 @@
...
+	if ply:PS_HasPoints(amount) and target:PS_HasPoints(amount) and CoinFlips[ply:SteamID()] == nil then
+		-- Both players have enough points AND an existing coinflip doesn't exist.
+		ply:ChatPrint("[orange]"..to:Nick().." is receiving your coinflip request.")
+		CoinFlips[ply:SteamID()] = {target:SteamID(), amount, CurTime()}
+	elseif CoinFlips[ply:SteamID()] ~= nil then
+		ply:ChatPrint("[red]What the hell r yeh doing bruv? yeh already hae a filp goin!")
+	elseif not ply:PS_HasPoints(amount) then
+		ply:ChatPrint("[red]What the hell r yeh doing bruv? yeh don't hae dooze points!")
+	elseif not target:PS_HasPoints(amount) then
+		-- Target doesn't have the required funds. Nothing happens, CoinFlipRequestTimeout will be called and timeout the roll.
```
swampservers: They can still infer from the "is recieving your coinflip request" thing. Just don't check the target's number of points but prohibit them from accepting it later if they don't have enough points.

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,153 @@
...
+	end
+end
+
+function CoinFlipRemove(id)
```
swampservers, line 63: cleanup unused

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,153 @@
...
+		-- Instead of taking the amount away from both and then giving the winner the amount x 2, simply remove/add here
+		if heads then
+			to:PS_TakePoints(amount)
+			from:PS_GivePoints(amount)
+		else
+			from:PS_TakePoints(amount)
+			to:PS_GivePoints(amount)
+		end
+	elseif not from:PS_HasPoints(amount) then
+		-- Initiator suddenly doesn't have the required funds. Nothing happens, show it off as an error.
```
swampservers: you don't need to worry about which of the two players doesn't have enough points, just throw the same error either way so the code is smaller

lua/chatcommands/coinflip.lua
```diff
@@ -0,0 +1,153 @@
...
+		elseif(fromPlayer ~= nil and toPlayer == nil) then
+			fromPlayer:ChatPrint("[fbc]The player you requested a coinflip to has left. Try again lates.")
+		end
+		table.remove(CoinFlips, usedID)
+	end
+end)
+
+function initCoinFlip(ply,target,amount)
+	if ply:PS_HasPoints(amount) and target:PS_HasPoints(amount) and CoinFlips[ply:SteamID()] == nil then
+		-- Both players have enough points AND an existing coinflip doesn't exist.
```
swampservers: I appreciate that you're trying to make quality code with all these comments, but actually some of these comments are obvious just by reading the code, so in the future you don't need to do them.

Ugleh: Some of the comments are also landmarks for my eyes when I need to go back to stuff and then I just leave them in.

swampservers:
I think this got merged in v3

PR #148, opened 2020-05-26
Changed Lock to Freeze, Modified ExtEmitSound's, added WorldModel to indicate who's "IT"

Ugleh:


lua/weapons/weapon_tag.lua
```diff
@@ -144,6 +144,32 @@ function SWEP:PrimaryAttack( right )
...
+function SWEP:CreateWorldModel()
+   if not IsValid(self.WModel) then
+      self.WModel = ClientsideModel(self.WorldModel,RENDERGROUP_OPAQUE)
+      self.WModel:SetNoDraw(true)
+      self.WModel:SetBodygroup(1,1)
+   end
+   return self.WModel
+end
+
+function SWEP:DrawWorldModel()
```
swampservers, line 156: this only renders when they have this as the active weapon, so they can hide it

Ugleh: I understand. But that can also be said about jihad weapon.

PR #155, opened 2020-07-13
Peacekeeper

Noztik:


lua/weapons/weapon_peacekeeper.lua
```diff
@@ -0,0 +1,341 @@
...
+
+	end
+end
+
+function SWEP:Initialize()
+	self:SetWeaponHoldType( self.HoldType )
+	self:SetHoldType( self.HoldType )
+end
+
+function SWEP:GetCone()
```
swampservers, line 142: Did you adjust the cone size from what the garand had? it should be big because its a shotgun. You can remove the sprint logic as well if you don't think its needed

Noztik: I did change the cone size

lua/weapons/weapon_peacekeeper.lua
```diff
@@ -103,24 +103,20 @@ function SWEP:Deploy()
 end
 
 function SWEP:SprintMod()
-	return ((math.cos(((self.SprintNess or 0)+1)*math.pi)+1)*0.5)
-end
-
-function SWEP:IsSprinting()
-	return self.Owner:GetVelocity():Length() > (self.Owner:GetWalkSpeed()+self.Owner:GetRunSpeed())*0.5
+	return ((math.cos((1)*math.pi)+1)*0.5)
```
swampservers: can you clean this up properly? if you have no sprinting behavior, there shouldn't be anything sprint-related in the file

Noztik: this is for weapon swaying while running

swampservers: it returns the same value no matter what though, it doesn't do anything. whatever it returns can just be a constant value where it is called. The weapon swaying while sprinting is a built-in thing.

Noztik: I can't believe I didn't notice that lol

lua/weapons/weapon_peacekeeper.lua
```diff
@@ -0,0 +1,310 @@
...
+		if (SERVER) then
+			self.Owner:SetFOV( 0, 0.15 )
+			--self:SetIronsights(false)
+		end
+	
+		if SERVER and self.Owner:Alive() then
+			local timerName = "ShotgunReload_" ..  self.Owner:UniqueID()
+			timer.Create(timerName,
+			(.5 + .05),
+			shellz,
```
cannon, line 264: Isn't this a double barrel shotgun where the reload animation reloads both shells? Why is this coded like the type of shotgun you insert shells one at a time?

swampservers, line 264: Isn't this a double barrel shotgun where the reload animation reloads both shells? Why is this coded like the type of shotgun you insert shells one at a time?

PR #157, opened 2020-08-09
ps_buy Change

chev2:
In the case a player buys a weapon from the pointshop using the ps_buy command and they already have that weapon, it'll just switch to that instead of wasting points


swampservers:
There are cases like the peacekeeper and previously the AR2 which are "WeaponAndAmmoProduct". In those cases, you'd want to be able to buy more ammo.

PR #159, opened 2020-08-10
AR2 fix, peacekeeper popup, remove old location code

chev2:
Fixed the AR2 angles & position

The 'you are in a safe space' popup will now show if you're holding the peacekeeper

Removed code for locations that aren't in the map e.g. crawl space and upper/lower caverns


swampservers:
Thank you!

PR #160, opened 2020-08-11
Coinflip Fix

Noztik:
wtf was with the chat messages in this lol


swampservers:
thanks for fixing the le funny messages

PR #162, opened 2020-08-12
Coinflip Fix 3

Noztik:
of course this guy even fucked up the loop despite using a consistent variable name for every loop iterating through this specific table


Noztik:
this is mostly what I just eyeballed and saw as an issue

PR #163, opened 2020-08-12
Coinflip Fix 4

Noztik:
hopefully this is the last fix for this, the rest of the system seems to be working as intended


swampservers:
thanks for fixing the mic

PR #166, opened 2020-08-19
locations update ?

ghost:
making clam and concessions attic appear on the TAB menu


swampservers:
why would you put a secret theater there...

PR #169, opened 2020-09-07
Taser SWEP

chev2:
A police taser.

Left click: If a player is in range, it attaches taser tongs to them. If the player gets out of range, or the player dies, or the attacker dies, etc. it removes the tongs.
Right click: If tongs are attached to a valid player, it will freeze them in place and make their playermodel go nuts. The player cannot switch weapons or look around. The attacker can tase/untase on command with right click, and the maximum time a player can be tased is 20 seconds.
Reload: Removes the tongs from an attached player, if there is one.

It should also properly remove tongs when using !drop (via OnRemove), and a tased player should go into third person/


swampservers:
Looks pretty good, let's merge and try it on the server. Thanks!

PR #170, opened 2020-09-09
Tasing fix

GrateBrittan:
Added a check to SWEP:Think() to make sure that people not being tased are untased. I was being tased earlier and when they stopped tasing me I was still in the tased mode should also fix people staying in the seizure after being untased

### DIFF
--- a/lua/weapons/weapon_taser.lua
+++ b/lua/weapons/weapon_taser.lua
@@ -106,11 +106,14 @@ function SWEP:Reload() -- remove the attachment
     self:RemoveWire()
 end
 
-function SWEP:Think() -- remove the rope if the player goes too far
+function SWEP:Think() -- remove the rope if the player goes too far and untases people not being tased
     if IsValid(self.AttachedPlayer) then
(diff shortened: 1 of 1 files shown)


lua/weapons/weapon_taser.lua
```diff
@@ -106,12 +106,15 @@ function SWEP:Reload() -- remove the attachment
...
 end
 
-function SWEP:Think() -- remove the rope if the player goes too far
+function SWEP:Think() -- remove the rope if the player goes too far and untases people not being tased
     if IsValid(self.AttachedPlayer) then
         if self.Owner:GetPos():DistToSqr(self.AttachedPlayer:GetPos()) > self.MaxHoldDist*self.MaxHoldDist or !self.AttachedPlayer:Alive() then
             self:RemoveWire()
         end
     end
+    if !IsValid(ply.IsTaseredBy) then
```
swampservers, line 115: ply is not defined in this context. It looks like UntasePlayer is already called by RemoveWire so I don't think this is the fix

GrateBrittan: Sorry, I'm new to writing lua in garry's mod. I'll try finding a proper fix later

PR #172, opened 2020-09-25
Fixed twitch

GrateBrittan:
added the LoadVideo function so it would actually send the video data to the cinema embed, sending swamp the html file for twitch so that the js is updated

### DIFF
--- a/gamemodes/cinema/gamemode/modules/theater/services/sh_twitch.lua
+++ b/gamemodes/cinema/gamemode/modules/theater/services/sh_twitch.lua
@@ -7,8 +7,8 @@ SERVICE.Name 	= "Twitch"
 SERVICE.NeedsCodecs = true
 SERVICE.NeedsChromium = true
 
-function SERVICE:GetKey( url )
-	if not string.match( url.host or "", "twitch.tv") then
+function SERVICE:GetKey(url)
+	if not string.match(url.host or "", "twitch.tv") then
 		return false
 	end
 
@@ -22,4 +22,13 @@ function SERVICE:GetKey( url )
 	return key
 end
 
-theater.RegisterService( 'twitch', SERVICE )
+if CLIENT then
+	function SERVICE:LoadVideo(Video, panel)
+		panel:EnsureURL("http://swampservers.net/cinema/"..self:GetClass()..".html")
+
+		local str = string.format("th_video('%s');", string.JavascriptSafe(Video:Key()))
+		panel:QueueJavascript(str)
+	end
+end
+
+theater.RegisterService('twitch', SERVICE)


swampservers:
Unless i'm missing something, adding this code makes no difference. The service tables inherit from sh_base, that's why its the baseclass. https://github.com/swampservers/contrib/blob/master/gamemodes/cinema/gamemode/modules/theater/sh_services.lua

Does twitch fail to request, or does it request and play but only show a black screen?

GrateBrittan:
> Unless i'm missing something, adding this code makes no difference. The service tables inherit from sh_base, that's why its the baseclass. https://github.com/swampservers/contrib/blob/master/gamemodes/cinema/gamemode/modules/theater/sh_services.lua
> 
> Does twitch fail to request, or does it request and play but only show a black screen?

At the moment it requests but nothing is sent to the embeded player, showing just a black screen. This fix would send the values needed for the video player to work, the html was also updated so that it works with the current twitch API

swampservers:
Yeah dude I know I wrote 90% of our custom cinema. Do you understand my comment?

DanTheBritish:
According to the [Twitch documentation](https://dev.twitch.tv/docs/embed/video-and-clips#interactive-frames-for-live-streams-and-vods) a channel is required when launching their JS embed player. I can't see on their changelog when this requirement was introduced.

To test if this was the issue, I loaded up the [Twitch HTML template](http://swampservers.net/cinema/twitch.html) and yep - there's an immediate error before th_video javascript can be called: `MissingParameterError: A channel, video, or collection id must be provided in options`

Should be a simple fix in JS to launch the player when th_video is first called, instead of using player.setChannel

DanTheBritish:
I got bored and wrote the JS anyway. Don't see the HTML file in this repo so will just dump here. This works with first launching a twitch channel and supports changing channel without reloading the whole frame.

```
<script>

  var player = null;
  var player_ready = false;
  var to_video = null;
  var to_volume = 100;
  var start_time = 0;
  //var is_live = false;

  var last_quality = "";

  function think() {
    if (player_ready) {
      var quality = window.innerHeight < 700 ? "medium" : "chunked";

      if (quality != last_quality) {
        player.setQuality(quality);
        last_quality = quality;

        // get rid of spinner
        player.pause();
        player.play();
      }

      if (to_video !== null) {
        player.setChannel(to_video);
        player.setQuality(quality);

        // get rid of spinner
        setTimeout(function() {
            player.pause();
            player.play();
        }, 4000);

        to_video = null;
      }

      if (to_volume !== null) {
        player.setVolume(to_volume/100);
        player.setMuted(to_volume==0);
        to_volume = null;
      }
    }
  }
  setInterval(think, 100);

  function th_video(id) {
    if (player == null) {
      player = new Twitch.Player("player", {
               height: "100%",
               width: "100%",
               channel: id
      });
      player_ready = true;
    }
    to_video = id;
  };

  function th_volume( volume ) {
    to_volume = volume;
  };

  function th_seek( seconds ) {
    console.log("can't seek");
  };
</script>

```

swampservers:
@DanTheBritish Thank you! I've put that on the server

DanTheBritish:
I just did a quick test and it's not worked - the player is no longer giving an error and twitch is returning a 200 success, but still not launching correctly in chromium. Debugging now.

DanTheBritish:
Ok I have confirmed the fix actually did work - however, the swamp HTML template now requires to be launched over HTTPS due to the Twitch changes. I tested on Swamp live and it doesn't matter if the client provides HTTP or HTTPS it always appears to default to HTTP. This is likely due to `panel:EnsureURL` specifying HTTP. 

I can make the change in the Lua file myself but can't do it in this PR, and I also can't test locally as I do not have source code. Which would you prefer @swampservers, leave it to me and create a new PR (no testing) or should I leave it to you and @Britain1944?

swampservers:
Thanks for looking into this for me! I've just updated twitch to use HTTPS.

DanTheBritish:
Confirmed that as of server restart earlier Twitch is now fixed.

GrateBrittan:
Closing pr due to the fix being implemented last night.

PR #173, opened 2020-09-28
Laser Pointer SWEP

GrateBrittan:
It's finally here, now stop bugging me about it.


lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
+-- This file is subject to copyright - contact swampservers@gmail.com for more information.
+
+SWEP.PrintName = "Laser Pointer"
+SWEP.Author = "Britain1944"
+SWEP.Instructions = "Left click: Turn on and off the laser.\nRight click: Change laser power\nRepurchase for more ammo."
```
swampservers: Can it be hold down to use the laser? you can use SWEP.Automatic = true and some kind of timer logic to disable it when the weapon hasn't been fired in the past 0.1 second

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+	else
+		self:SetNWBool("CanDrawLaser",false)
+	end
+end
+
+function SWEP:BlindPlayer(ply)--makes the player being aimed at go blind
+	if SERVER then
+		local blind = ply:GetNWFloat("Blindness", 0)
+		blind = (blind or 0)+(1+self.LaserPower)*15
+		if (blind or 0) > 0 then
```
swampservers: remove "or 0", use `math.Clamp`

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+		self.Owner:RemoveAmmo(100-self.Weapon:Clip1(), self.Primary.Ammo, false)
+		self.Weapon:SetClip1(100)
+		self.Owner:EmitSound("items/ammo_pickup.wav")
+		timer.Remove("Reload"..self.Owner:SteamID64())
+	end)
+end
+
+function SWEP:Think()
+	if self:Clip1() > 0 then
+		self:SetNWBool("CanDrawLaser",true)
```
swampservers: `self:SetNWBool("CanDrawLaser",self:Clip1() > 0)`

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+	local loc = Location.GetLocationNameByIndex(Location.Find(self.Owner)):lower()
+
+	if (self.Owner:InTheater() and not (self.Owner:GetTheater()._AllowItems)) or loc=="trump lobby" or loc=="golf" then
+		self.Owner:PrintMessage(HUD_PRINTTALK, "[red] Stop right there! Lasers are not allowed in this area. ;authority;")
+	end
+
+	self:DrawLaser()
+	self.Weapon:TakePrimaryAmmo(self.LaserPower/2+0.1)
+
+	if IsValid(trc.Entity) and trc.Entity:IsPlayer() and trc.HitBox == 0 and not -Safe(trc.Entity) and (trc.Entity:InTheater() and not (trc.Entity:GetTheater()._AllowItems)) or lctn=="trump lobby" or lctn=="golf" then
```
swampservers: why is there a dash `-Safe` ?

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+		self:BlindPlayer(trc.Entity)
+		self:SetNextPrimaryFire(CurTime() + 0.1)
+	else
+		self:SetNextPrimaryFire(CurTime() + 0.1)
```
swampservers: factor out of conditional

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+
+function SWEP:PrimaryAttack()--shoots the laser
+	if (!self:CanPrimaryAttack()) then return end
+
+	local trc = self.Owner:GetEyeTrace()
+	local lctn = Location.GetLocationNameByIndex(Location.Find(trc.Entity)):lower()
+	local loc = Location.GetLocationNameByIndex(Location.Find(self.Owner)):lower()
+
+	if (self.Owner:InTheater() and not (self.Owner:GetTheater()._AllowItems)) or loc=="trump lobby" or loc=="golf" then
+		self.Owner:PrintMessage(HUD_PRINTTALK, "[red] Stop right there! Lasers are not allowed in this area. ;authority;")
```
swampservers: With the toggle system can't you just turn it on outside a theater and walk in?

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+		local tr = self.Owner:GetEyeTrace()
+		local vm = self.Owner:GetViewModel()
+		local attachmentIndex = vm:LookupAttachment("tip")
+		local origin = vm:GetAttachment(attachmentIndex).Pos
+		local lsrclr = self.LaserColors[self.LaserPower+1]
+
+		cam.Start3D(EyePos(), EyeAngles())
+			local size = math.random()*10
+
+			render.SetMaterial(Material("sprites/bluelaser1"))
```
swampservers: Factor out  a DrawLaser function with origin and direction as arguments

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+    end)
+    timer.Create("BlindnessUpdater", 0.1, 0, function()
+    	for i,v in next, player:GetAll() do
+    		if IsValid(v) then
+    		local blind = v:GetNWFloat("Blindness", 0)
+				if blind ~= nil then
+					if (blind or 0) > 0 then blind = blind-2 end
+					if blind < 0 then blind = 0 end
+
+					v:SetNWFloat("Blindness",blind)
```
swampservers: Don't call this unless the value has changed, i'm not sure if it networks unchanged values

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+	hook.Add("DoPlayerDeath", "OnPlayerDeath", function(ply, att, dmg)
+        if IsValid(ply:GetActiveWeapon()) and ply:GetActiveWeapon():GetClass() == "weapon_laserpointer" then
+            ply:GetActiveWeapon():RemoveLaser()
+        end
+    end)
+    timer.Create("BlindnessUpdater", 0.1, 0, function()
+    	for i,v in next, player:GetAll() do
+    		if IsValid(v) then
+    		local blind = v:GetNWFloat("Blindness", 0)
+				if blind ~= nil then
```
swampservers: It's not nil because you set a default

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+						render.SetMaterial(Material("sprites/light_glow02_add_noz"))
+						render.DrawSprite(trc.HitPos, size, size, lsrclr)
+
+						maxbounce = maxbounce+1
+					end]]
+				cam.End3D()
+			end
+		end
+	end)
+	hook.Add("KeyRelease","RemoveBeam", function(ply, key)
```
swampservers: This copy pasted function should just be outside the `if CLIENT then ... else ... end`

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+	end)
+	hook.Add("HUDPaintBackground", "BlindnessDraw", function()
+		local ply = LocalPlayer()
+		if IsValid(ply) then
+			surface.SetDrawColor(255,255,255,ply:GetNWFloat("Blindness",0))
+			surface.DrawRect(0, 0, ScrW(), ScrH())
+		end
+	end)
+	hook.Add("PostDrawTranslucentRenderables", "DrawAllLasers", function()
+		for i,v in next, player.GetAll() do
```
swampservers: Unfortunately player.GetAll() returns a newly made table and it's not acceptable to use this function in a drawing hook like this because it's too expensive.

What you need to do is create a global table containing all the players with laser pointers that are turned on, and update that table whenever it changes

lua/weapons/weapon_laserpointer.lua
swampservers: SetNWFloat also will send the value to every player on the server, so it would be better to use a custom net message of some sort only send to the player in question

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,326 @@
...
+		if IsValid(wep) and wep:GetClass() == "weapon_laserpointer" then
+			draw.RoundedBox(8, (ScrW()/2)-64, ScrH()-48, 128, 32, Color(0, 0, 0, 128))
+			draw.RoundedBox(0, (ScrW()/2)-58, ScrH()-42, 1.16*wep:Clip1(), 20, wep.LaserColors[wep.LaserPower+1])
+			draw.DrawText("Battery", "DermaLarge", (ScrW()/2), ScrH()-84, Color(255, 255, 255, 255), TEXT_ALIGN_CENTER)
+		end
+	end)
+	hook.Add("HUDPaintBackground", "BlindnessDraw", function()
+		local ply = LocalPlayer()
+		if IsValid(ply) then
+			surface.SetDrawColor(255,255,255,ply:GetNWFloat("Blindness",0))
```
swampservers: This seems kind of lame, it just whites out your screen?

Can you make it draw a flare at the :ToScreen() of the laser source or something? https://pcmedia.gamespy.com/pc/image/article/118/1184679/bf3lasersight_1311840613.jpg

lua/weapons/weapon_laserpointer.lua
GrateBrittan: Yes the weapon is an automatic weapon, so it turns on when you shoot and turns off when you're not shooting

lua/weapons/weapon_laserpointer.lua
GrateBrittan: Will do

lua/weapons/weapon_laserpointer.lua
GrateBrittan: okay thanks for helping me condense!

lua/weapons/weapon_laserpointer.lua
GrateBrittan: oh shit oops, that was originally commented for testing guess I forgot to remove the second -

lua/weapons/weapon_laserpointer.lua
GrateBrittan: you can but once you let go you can't use it anymore, and it checks that you're not blinding a player in the safe/banned areas

lua/weapons/weapon_laserpointer.lua
GrateBrittan: okay

lua/weapons/weapon_laserpointer.lua
GrateBrittan: during testing it kept on saying that it was nil and causing errors, that's just there to not cause the errros

lua/weapons/weapon_laserpointer.lua
GrateBrittan: okay, that's actually a good idea

lua/weapons/weapon_laserpointer.lua
GrateBrittan: oh okay

lua/weapons/weapon_laserpointer.lua
GrateBrittan: maybe, I was just wanted the effect to be like a flashbang

lua/weapons/weapon_laserpointer.lua
GrateBrittan: does this mean you want me to move the SetNextPrimaryFire out of the conditional?

lua/weapons/weapon_laserpointer.lua
GrateBrittan: I don't know about this one, I wanted the laser to act as similar to the physgun beam and right now it does. If I I make that change then it'll change it from the original idea I had, but if you want it that way I will change it.

lua/weapons/weapon_laserpointer.lua
GrateBrittan: but I don't think the weapon needs the second part

lua/weapons/weapon_laserpointer.lua
GrateBrittan: but you can also do the same with vapes, I've done it with both the mega and dragon breath vape.

lua/weapons/weapon_laserpointer.lua
swampservers: why would it act different? I mean just take this code that is copied in 3 different locations and put it in one function, then call that function

lua/weapons/weapon_laserpointer.lua
swampservers: try removing it with a fresh server

lua/weapons/weapon_laserpointer.lua
swampservers: yeah there is no reason to have it on both side of if/else, just put it at the bottom

lua/weapons/weapon_laserpointer.lua
swampservers: the instructions imply it's a toggle to turn it on and off, i think it should be you hold it down to keep it on

lua/weapons/weapon_laserpointer.lua
swampservers: plus this way the target will see where the blinding laser is coming from so they can go kill the guy doing it, just making the screen white is just annoying. you can make the flare much bigger and more blinding than the image i linked, that is just to show the idea

the GetNWFloat("Blindness") thing won't give you enough info to achieve this, but you don't need it, you can implement the BlindnessDraw function using the same global table of turned on laser pointers that you use for the beam drawing. If you have any fade in/out, that should be implemented clientside anyway, so the fading isn't laggy and updated at 15-25fps (the server tickrate)

lua/weapons/weapon_laserpointer.lua
GrateBrittan: Yeah lemme change the instructions, it doesn't say toggle but it doesn't say that you hold it down for laser

lua/weapons/weapon_laserpointer.lua
GrateBrittan: Yep already done

lua/weapons/weapon_laserpointer.lua
GrateBrittan: Will do

lua/weapons/weapon_laserpointer.lua
GrateBrittan: Okay

lua/weapons/weapon_laserpointer.lua
```diff
@@ -0,0 +1,270 @@
...
+		local wep = LocalPlayer():GetActiveWeapon()
+		if IsValid(wep) and wep:GetClass() == "weapon_laserpointer" then
+			draw.RoundedBox(8, (ScrW()/2)-64, ScrH()-48, 128, 32, Color(0, 0, 0, 128))
+			draw.RoundedBox(0, (ScrW()/2)-58, ScrH()-42, 1.16*wep:Clip1(), 20, wep.LaserColors[wep.LaserPower+1])
+			draw.DrawText("Battery", "DermaLarge", (ScrW()/2), ScrH()-84, Color(255, 255, 255, 255), TEXT_ALIGN_CENTER)
+		end
+	end)
+	hook.Add("HUDPaintBackground", "BlindnessDraw", function()
+		local ply = LocalPlayer()
+		if IsValid(ply) then
```
swampservers: Try this
```
        for ent,_ in next, ON_LASERS do
            if not IsValid(ent) then ON_LASERS[ent] = nil continue end
            if ent.Owner==LocalPlayer() then continue end
            pos, ang, dist = ent:GetBeam()
            dir = ang:Forward()
            if (pos + dir*dist):Distance(EyePos()) < 20 then
                alpha = (2-dir:Dot(EyeAngles():Forward()))/2
                alpha = math.pow(alpha, 0.5)
                surface.SetDrawColor(255,255,255,alpha*255)
                surface.SetMaterial( FLARE SPRITE HERE )
                sp = pos:ToScreen()
                local sz = ScrH()
                if sp.visible then
                    surface.DrawTexturedRect(sp.x - sz/2,sp.y - sz/2,sz,sz)
                end
            end
        end
```

And add these functions:
```
function SWEP:HandPos()
    if self.Owner == LocalPlayer() and LocalPlayer():EyePos():Distance(EyePos()) < 1 then
        pos,ang = EyePos(), EyeAngles()
        pos = pos + ang:Right()*2
        pos = pos + ang:Up()*-3
        pos = pos + ang:Forward()*5
        ang:RotateAroundAxis(ang:Forward(), -3)
        ang:RotateAroundAxis(ang:Right(), -90)
        ang:RotateAroundAxis(ang:Up(), -105)
        pos = pos + ang:Forward()*0.1
        return pos
    end

    local bn = "ValveBiped.Bip01_R_Hand"
    local bon = ply:LookupBone(bn) or 0

    local opos = self:GetPos()
    local oang = self:GetAngles()
    local bp,ba = ply:GetBonePosition(bon)
    if(bp)then opos = bp end
    if(ba)then oang = ba end

    oang:RotateAroundAxis(oang:Right(), -90)
    oang:RotateAroundAxis(oang:Up(), 90)
    opos = opos + oang:Forward()*-2
    opos = opos + oang:Right()*1.25
    opos = opos + oang:Up()*1.5
    return opos
end

-- use the first two returns for DrawWorldModel and GetViewModelPosition as well
function SWEP:GetBeam()
    local endpos = self.Owner:EyePos() + self.Owner:EyeAngles():Forward() * 10000
    local tr = util.TraceLine( {
        start = self.Owner:EyePos(),
        endpos = endpos,
    } )

    if tr.Hit then
        endpos = tr.HitPos
    end

    local orgpos = self:HandPos()
    local dir = (endpos - handpos)
    return handpos, dir:GetNormalized():Angle(), dir:Length()
end```

ContagiousBR: VENT CLOGGER

PR #175, opened 2020-10-09
DLive Support

Noztik:


gamemodes/cinema/gamemode/modules/theater/services/sh_dlive.lua
```diff
@@ -0,0 +1,82 @@
+-- This file is subject to copyright - contact swampservers@gmail.com for more information.
+
+local SERVICE = {}
+
+SERVICE.Name 		= "DLive"
```
GrateBrittan, line 5: Why the hell is there a large gap between SERVICE.Name and the equal sign?

swampservers: it was originally so that the equals signs line up vertically between different member variables (some with longer names), if that's not the case here it was just copied

gamemodes/cinema/gamemode/modules/theater/services/sh_dlive.lua
```diff
@@ -0,0 +1,82 @@
...
+
+function SERVICE:GetKey(url)
+	if url.host and string.match(url.host,"dlive.tv") and string.match(url.path,"^/([%w_]+)[/]?$") then
+		return url.encoded
+	end
+	return false
+end
+
+if CLIENT then
+	function SERVICE:GetVideoInfoClientside(key, callback)
```
swampservers, line 17: It's probably possible to get the title with serverside HTTP, but I don't see it anywhere in the initial HTML page so this is fine

swampservers:
Thanks! Let's test it

PR #179, opened 2020-10-14
Dlive Fix 4

Noztik:
non-hd and hd have different screen sizes and the non-hd size puts dlive into a mobile view mode


swampservers:
FWIW, I think the non-hd mobile 360p streams load more consistently (dlive's CDN isn't the best)

PR #180, opened 2020-10-14
Dlive Fix 5

Noztik:


gamemodes/cinema/gamemode/modules/theater/services/sh_base.lua
```diff
@@ -90,8 +90,10 @@ if CLIENT then
 	end
 
 	function SERVICE:SetVolume(vol, panel)
-		local str = string.format("th_volume(%s);", vol)
-		panel:QueueJavascript( str )
```
swampservers, line 94: Are you sure this is an issue? Maybe it has something to do with being redirected and the JS you're trying to execute runs before the redirection resets the page?

Noztik: I don't understand what you're trying to say about redirection, SetVolume gets run when loading the video and when you adjust the slider. I have hijacked the SetVolume function ingame and it works fine when I swap QueueJavascript for RunJavascript.

PR #190, opened 2020-11-10
Laser fixes

PYROTEKNIK:
some fixes that don't appear to have made it on


lua/sps/items/other.lua
```diff
@@ -58,6 +58,15 @@ PS_WeaponProduct({
 	class = 'weapon_taser'
 })
 
+PS_WeaponProduct({
+	name = 'Laser Pointer',
+	description = "Makes a funny dot. Keep away from eyes. Right click for lethal beam. Re-buy for battery refill.",
+	price = 350,
```
GrateBrittan: Price should be higher, current price is wayyy to overpowered. This is cheapest weapon on the shop, I think it should be like 1000 or more.

swampservers: Isn't it already buyable through console and therefore already being registered somewhere? also yeah too cheap

PYROTEKNIK: I moved it to the file that contained other items in the expensive toys category. Also I feel the price is justified since its capacity to kill is basically just above nothing. One battery has the potential to do exactly 250 damage to one person at a time over a period of 50 seconds. Please provide a price suggestion if you do not agree with mine.

PR #193, opened 2020-11-15
More Reliable Service Switching

Noztik:


gamemodes/cinema/gamemode/modules/scoreboard/controls/cl_html.lua
```diff
@@ -131,7 +131,8 @@ function PANEL:Think()
 		-- Run queued javascript
 		if self.JS then
 			for k, v in pairs( self.JS ) do
-				self:RunJavascript( v )
+				if (LocalPlayer().videoDebug) then print("(JS)",v) end
+				self:RunJavascript( string.format( "window.addEventListener('load',function(){%s});if(document.readyState=='complete'){%s};", v, v ) )
```
swampservers: Can you only add the event listener in an "else" condition on document.readyState?

Noztik: If the page is loading, document.readyState will be "loading" and the load event will activate when the page finishes loading. If the page is loaded, the load event listener will do nothing and document.readyState will be "complete" so there is no need for an "else" condition. Only one can activate per page state.

swampservers: what I mean is, if the page is already loaded when QueueJavascript is called, this will add event listeners that never get run, so only add the event listener if it's not loaded.

there is also an issue here with the event listeners potentially being called out of order when the page loads if you QueueJavascript multiple times, a fix for that would be to only create one event listener and then put all the queued javascript in a list of anonymous functions

Noztik: I don't know what you mean by "list of anonymous functions"

swampservers: An "anonymous function" is an unnamed function that is treated like a value

Try like this:

```
if(document.readyState=='complete'){
%s;
} else {

if (queuedJS === undefined) {
var queuedJS=[];

window.addEventListener('load',function(){
for (i in queuedJS){
queuedJS[i]();
}
});

}

queuedJS.push(function(){%s;});
}
```

Also test it to make sure they run in order, it's possible that the queuedJS thing in this code will have a scope issue and not share its value between each QueueJavascript call.

NottyNoz: how would I only send 1 event listener per page when QueueJavascript can be run at any time after loading the page

PR #203, opened 2021-02-04
HLS Rework

Noztik:
I realized that file thumbnails could have explicit material in them however there aren't any built-in ways to get the maturity rating of a video in a theater from outside the theater so it had to be hardcoded like this.

### DIFF
--- a/gamemodes/cinema/gamemode/modules/theater/ents/entities/theater_thumbnail/cl_init.lua
+++ b/gamemodes/cinema/gamemode/modules/theater/ents/entities/theater_thumbnail/cl_init.lua
@@ -144,7 +144,7 @@ end
 function ENT:DrawThumbnail()
 
 	-- Thumbnail isn't set yet
-	if self:GetThumbnail() == "" or (self:GetService() == "file" and not GetConVar("swamp_mature_content"):GetBool()) then
+	if self:GetThumbnail() == "" or (theater.Services[self:GetService()] and theater.Services[self:GetService()].Mature) then
 
 		if self:GetService() == "" then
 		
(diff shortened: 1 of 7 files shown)


gamemodes/cinema/gamemode/modules/theater/ents/entities/theater_thumbnail/cl_init.lua
```diff
@@ -144,7 +144,7 @@ end
 function ENT:DrawThumbnail()
 
 	-- Thumbnail isn't set yet
-	if self:GetThumbnail() == "" then
+	if self:GetThumbnail() == "" or (self:GetService() == "file" and not GetConVar("swamp_mature_content"):GetBool()) then
```
swampservers: `theater.Services[self:GetService()].Mature` should work (from gamemodes/cinema/gamemode/modules/theater/sh_services.lua, RegisterService())

Noztik:
please change hls.html to https://pastebin.com/YcYNbW43

gamemodes/cinema/gamemode/modules/theater/services/sh_dlive.lua
```diff
@@ -15,93 +15,17 @@ function SERVICE:GetKey(url)
 end
```
swampservers, line 16: I don't think this service needs to be maintained since no one uses it anymore, but it's up to you

Noztik: yea I know but I already wrote it

swampservers:
Thank you for continuing to work on this stuff; I sent you some paypal$

PR #205, opened 2021-02-08
Pointshop Freeze Fix

Noztik:


swampservers:
I think a simpler solution for the pointshop-freeze would be to just self:SetModel() inside the Paint function for the DModelPanel or its parent. The weird thing though, is that I can't reproduce the original issue anymore; the game doesn't freeze for me. Does it still freeze for you?

Noztik:
It still noticeably freezes for me.

swampservers:
@NottyNoz I just did a really simple fix and put it on the server, can you please check if the freezing issue is still there? I didn't think of this fix earlier

Noztik:
Your fix is as good as the fix I wrote. I think my fix is simpler than yours.

swampservers:
lolwut? this is my fix https://github.com/swampservers/contrib/commit/807260e89a544d781f9981831b6e082305ba2d60
nothing else i committed is relevant to this issue, nor is the moving of PS_PreRender

Noztik:
My fix that I committed before yours does just as good a job at reducing the lag as your new fix does.

swampservers:
Yeah but my fix is like 3 statements in one file

gamemodes/cinema/gamemode/modules/theater/services/sh_file.lua
```diff
@@ -62,6 +62,9 @@ if CLIENT then
 
 					if duration>0 then
 						callback({duration=duration})
+					elseif duration<2147483647 then
```
swampservers, line 65: this if/elseif/else block doesn't make sense, what values do you expect for these cases?

Noztik: Hacked/corrupted files return large durations so if the duration is over the 32 bit limit then it's probably a hacked file.

NottyNoz: > Yeah but my fix is like 3 statements in one file
mine is a single if statement lol

Noztik:
> Yeah but my fix is like 3 statements in one file

mine is a single if statement lol

gamemodes/cinema/gamemode/modules/theater/services/sh_file.lua
swampservers: you're using < not >

swampservers:
Oh, yeah, I see you updated it now, my bad. Only problem I have with yours is that the `:GetChild(0)` thing would break if you added other sub-panels to the "item" (eg, if the text over it was changed to be a panel rather than just painting it). I think mine makes it a little more clear the way "model2set" gets set on initialization rather than the model itself, so it's clear that it's intentional not to set the model directly. It doesn't matter though, mine's already committed but i'll merge this anyway since it won't break anything

swampservers:
(you probalby want to fix the `< not >` thing later)

Noztik:
what `< not >` thing? are you talking about the mature boolean? it's set to true when you have the filter off

swampservers:
> what `< not >` thing? are you talking about the mature boolean? it's set to true when you have the filter off

the `elseif duration<2147483647 then`

PR #207, opened 2021-02-10
Streamwatch Fix

Noztik:
I want to have the option of choosing whether a video or service has to have the title fetched instead of grabbed from the logs like every request that has a duration of 0. HLS vods have to have their titles manually entered but their duration isn't 0 so the person who queues that link gets to choose its name for the future.


swampservers:
You can just set `SERVICE.CacheLife` (defaults to 7 days)

PR #208, opened 2021-02-10
HLS CORS Bypass

Noztik:
I could completely remove the clientside system to get the streamwat.ch hls link but I'll leave it in since people use streamwat.ch sort of often so a failsafe would be fine to keep in.


gamemodes/cinema/gamemode/modules/theater/services/sh_hls.lua
```diff
@@ -70,15 +72,15 @@ if CLIENT then
 	
 	function SERVICE:LoadVideo( Video, panel )
 		local k = Video:Key()
-		if (string.len(Video:Data())>1 and Video:Data() != "true") then
+		if (not isbool(Video:Data()) and string.len(Video:Data())>1) then
```
swampservers: Can we please keep the Data type as a string to prevent potentially breaking other code?

PR #211, opened 2021-02-14
ThirdParty Placeholder

Noztik:
I really just need the sv file so I can test and develop it on the server


swampservers:
Is this supposed to use your video-finder script to get the file link? Can you just do the link lookup before requesting the video on the server, so it requests as a file?

Noztik:
This is for playing videos that are tied to a site and must be played on their site/with their referer. You can't just play the video file on your own site. I'm probably going to have to add something to my video-finder script or make a new gui thing to queue two links at once in a json string table.

swampservers:
Oh, I see. So you need two links, one to the page, and one to the video right? Then you replace the page's HTML with our player and run javascript? You can use the Key() and Data() fields of the video together to make that work

Noztik:
Correct but I still need to request both links which means putting them in a string table.

PR #212, opened 2021-02-15
monkey swep + hints fix?

PYROTEKNIK:
i'm not sure why it's ignoring the convar. maybe this will help?
also there was a duplicate hook in the laser swep for some reason


lua/autorun/gamehints.lua
```diff
@@ -7,7 +7,7 @@ SwampHints["shop"] = "Press F3 to access the pointshop"
 SwampHints["disablehints"] = "To disable hints, open the scoreboard using tab and check the Display section"
 
 function SwampHint_InitVar()
```
swampservers, line 9: I'm pretty sure you can just do these like in https://github.com/swampservers/contrib/blob/master/gamemodes/cinema/gamemode/modules/theater/cl_commands.lua

swampservers:
Thank you!

lua/autorun/gamehints.lua
PYROTEKNIK: Well when it was accessing it as a var someone complained that it wasn't removing some of the hints.

PR #214, opened 2021-02-15
Update weapon_monke.lua

PYROTEKNIK:
i forgot to put in the most recent version of this before i did the original pr.
i did some indenting too


swampservers:
Maybe you can make it so you can like climb on the walls? If that is possible without making too many laggy hooks

PYROTEKNIK:
i could look into it

PR #219, opened 2021-02-20
Changes and Fixes

Noztik:
go to the `PrivateTheaterLightSwitch` hook in your cinema location file and change
`if Theater and Theater:IsPrivate() and ent:GetClass() == "func_button" then` to
`if Theater and Theater:IsPrivate() and ent:GetClass() == "func_button" and ent:GetPos().z != 11392 then`
this is to fix an issue where someone can put down a mobile theater at the moon base and completely prevent anyone except the owner from using the airlock doors


swampservers:
Sorry, was preoccupied IRL

PR #222, opened 2021-03-07
New Thumbnail Networking

Noztik:


swampservers:
Is this just to deal with the 255 char limit?

I'm worried this may not work because entities aren't Valid on the clientside until the client can "see" them in PVS

Noztik:
> 
> 
> Is this just to deal with the 255 char limit?
> 
> I'm worried this may not work because entities aren't Valid on the clientside until the client can "see" them in PVS

I've tested this on my own dedicated server and it doesn't throw any errors and works fine although I didn't really test the broadcast part when the entity wasn't in PVS
they should be valid regardless if they are in PVS or not anyway since the client can fetch them with ents.FindByClass("theater_thumbnail") anywhere

swampservers:
> > Is this just to deal with the 255 char limit?
> > I'm worried this may not work because entities aren't Valid on the clientside until the client can "see" them in PVS
> 
> I've tested this on my own dedicated server and it doesn't throw any errors and works fine although I didn't really test the broadcast part when the entity wasn't in PVS
> they should be valid regardless if they are in PVS or not anyway since the client can fetch them with ents.FindByClass("theater_thumbnail") anywhere

that's not true, if the entity is out of PVS it won't be in the returned table before the client has seen it for the first time. please do a bit of testing to learn about how entity networking works

gamemodes/cinema/gamemode/modules/theater/sh_theater.lua
```diff
@@ -283,8 +290,22 @@ if SERVER then
...
+			net.WriteEntity(self._ThumbEnt)
+			net.WriteString(self:VideoThumbnail())
+		net.Broadcast()
+		
+		ThumbnailTable = ThumbnailTable or {}
+		net.Receive("ThumbnailDelivery", function(len, ply)
+			local e = net.ReadEntity()
+			net.Start("ThumbnailDelivery")
+				net.WriteEntity(e)
+				net.WriteString(ThumbnailTable[e] or "")
```
swampservers: Should use entity.thumbnail on serverside too

PR #223, opened 2021-03-31
More button stuff, new dodgeballs, new spraypaint, kleiner revision

PYROTEKNIK:
Current Issues: 
Kleiners can sometimes become invincible, not sure of cause
Kleiner pathfinding could be more efficient

Please make sure you look over all this stuff i don't wanna break the server


PYROTEKNIK:
oh yeah you might want to tweak that MoneyPrize function in the magic buttons to something else if the prize value is too high

lua/entities/kleiner/cl_init.lua
```diff
@@ -0,0 +1,142 @@
...
+function ENT:Initialize()
+		local effectdata = EffectData()
+	effectdata:SetOrigin( self:GetPos() )
+	effectdata:SetEntity( self )
+	util.Effect( "propspawn", effectdata )
+end
+ 
+
+local spkr = Material("voice/icntlk_pl")
+hook.Add("CreateClientsideRagdoll","KleinerRagdollsFix",function(ent,rag)
```
swampservers: Does the ragdoll color not work otherwise?

lua/entities/kleiner/cl_init.lua
```diff
@@ -0,0 +1,142 @@
...
+		function rag:GetPlayerColor()
+			if(self.wasbased)then
+				return Vector(0.8,0,0.05)
+			end
+			return Vector(0.23,0.35,0.41)
+		end
+		
+		
+		if(IsValid(rag))then rag:SetSubMaterial(5,ent:GetSubMaterial(5,"")) end
+		timer.Create(ent:EntIndex().."ragcleaner",10,1,function()
```
swampservers: timer.Simple

lua/entities/kleiner/cl_init.lua
```diff
@@ -0,0 +1,142 @@
...
+	local target = self:GetTarget()
+
+	if(IsValid(target))then 
+		return target:EyePos()
+	end
+	return self:EyePos() + self:GetAngles():Forward()*600
+end
+
+
+function ENT:Think()
```
swampservers: Can this go into a draw hook, so it's not run all the time? Even if you have to call SetupBones

lua/entities/kleiner/cl_init.lua
PYROTEKNIK: GetPlayerColor only exists on player entity, so the material proxy typically only returns the default blue color unless you add that function to an entity.

lua/entities/kleiner/cl_init.lua
```diff
@@ -0,0 +1,142 @@
...
+	if(self:GetTalking() > CurTime())then
+		local bone = self:GetBonePosition(self:LookupBone("ValveBiped.Bip01_Head1") or 0)
+		render.SetMaterial( spkr )
+		render.DrawSprite( bone + Vector(0,0,16), 12, 12, color_white)
+	end
+end
+
+function ENT:GetPlayerColor()
+	if(self:GetBased())then
+	return Vector(0.8,0,0.05)
```
swampservers: can these colors (since they are copied in the ragdoll) be put in a single location?

lua/autorun/client/big_chungus_in_the_bathroom.lua
```diff
@@ -0,0 +1,34 @@
+BIG_CHUNGUS_LOCATION = Vector(0, 1598, 64)
```
swampservers: add the header

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,970 @@
...
+	self:DropGrenades()
+	
+	hook.Run( "OnNPCKilled", self, dmginfo:GetAttacker(), dmginfo:GetInflictor() )
+	if(SERVER)then
+		local attacker = dmginfo:GetAttacker()
+		
+		if(IsValid(attacker) and attacker:IsPlayer())then 
+		self:AddTargetViolence(attacker)
+			local reward = self:GetBased() and self.KillRewardBased or self.KillReward
+			if(attacker.PS_AddPoints)then attacker:PS_AddPoints(reward) end
```
swampservers: PS_GivePoints

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,970 @@
...
+end
+ 
+function ENT:CanBeTarget(ent)
+	if(!IsValid(ent))then return false end
+	if(ent == self)then return false end
+
+	if(ent.InVehicle and ent:InVehicle())then return false end
+	if(ent.Alive and !ent:Alive())then return false end
+	if(ent:GetMoveType() == MOVETYPE_FLY)then return false end
+	if(player.IsAFK and player:IsAFK())then return false end
```
swampservers: mixing `ent` and `player`. Do you expect this to target entities other than players? If not, don't add extra checks

PYROTEKNIK: they were designed originally to target eachother, but i figured it would be a waste of serverside logic to have them behave away from players. i will remove them

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,970 @@
...
+
+function ENT:HaveTarget()
+	local target = self:GetTarget()
+	if ( IsValid(target) and self:CanBeTarget(target) ) then
+		return true 
+	end
+	return self:FindTarget()
+end
+
+function ENT:GetTargetTable()
```
swampservers: Where is this used? Is there a reason you can't just use `KLEINER_NPC_TARGETS or {}` ?

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,970 @@
...
+	for key,ent in pairs(targets)do
+		samplevalue = samplevalue - self:GetTargetPriority(ent)
+		if(samplevalue <= 0)then
+		self:SetTarget(ent)
+		return true
+		end
+	end
+
+	if(playersum == 0)then --slowly die if they end up somewhere where nobody is
+		local dmg = DamageInfo()
```
swampservers: SetHealth?

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,970 @@
...
+		-- At this point in the code the bot has stopped chasing the player or finished walking to a random spot
+		-- Using this next function we are going to wait 2 seconds until we go ahead and repeat it 
+		coroutine.wait(0.4)
+		
+	end
+
+end
+
+
+function ENT:MoveToPos( pos, options )
```
swampservers: wait, isn't this method built into the NextBot?

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,970 @@
...
+			end
+		end	
+		if(start > 1)then start = start - 1 end
+		return self.path:GetAllSegments()[start],start
+	end
+	return nil,-1
+end
+
+function ENT:GetNextPathPoint(ahead)
+if(!self.path or !self.path:IsValid())then return end
```
swampservers: can you run a formatter please

lua/entities/kleiner/init.lua
PYROTEKNIK: i think i was originally using this in place of KLEINER_NPC_TARGETS before i had that table. i will remove this

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,970 @@
...
+		if(across)then
+			self.loco:JumpAcrossGap((seg3 and seg3.pos) or (seg2 and seg2.pos),dir)
+			
+		else
+			self.loco:Jump() 
+		end
+	end
+	
+	--NOTE: door use seems to be pretty expensive due to the tracing, i would like to optimize this so that it doesn't run unless the path contains a doorway.
+	--Disabling the following function will just make them get stuck and teleport past the door instead
```
swampservers: teleporting is fine

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,970 @@
...
+			return -1
+		end
+	
+		if ( !self.loco:IsAreaTraversable( area ) and !ladder:IsValid()) then
+			// our locomotor says we can't move here
+			return -1
+		end
+		
+		--if(!self:PosInRange(area:GetCenter()))then return -1 end
+		// compute distance traveled along path so far
```
swampservers: please do not use // comments in glua, it messes up the formatting in just about every editor

lua/entities/kleiner/init.lua
PYROTEKNIK: i had to make an alternate implementation of it that runs  self:WhilePathing(), which allows movement to be interrupted for handling things like doors

lua/entities/kleiner/init.lua
PYROTEKNIK: whoops that's leftover from the original function example on the wiki

lua/entities/kleiner/sv_spawning.lua
```diff
@@ -0,0 +1,62 @@
...
+		KLEINER_NPCS_CURRENT_NUMBER = table.Count(KLEINER_NPCS)
+	end	
+	
+end)
+hook.Add("OnEntityCreated","kleiner_npc_register",function(ent)
+	if(ent:GetClass() == "kleiner" or ent:IsPlayer())then
+		KLEINER_NPCS[ent] = ent:GetClass() == "kleiner" and true or nil
+		KLEINER_NPC_TARGETS[ent] =  ent:IsPlayer() and true or nil
+		
+		KLEINER_NPCS_FILTER = {}
```
swampservers: Remove this, use a function for the trace filter instead (if you were going to use it)

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,457 @@
...
+			tr.mins = hull*Vector(1,1,0.1)*-0.5
+			tr.maxs = hull*Vector(1,1,0.1)*0.5
+			trace = util.TraceHull(tr)
+			 
+			 
+			if((!self.PlacementSettings.RequireCeiling or (trace.Hit and self:IsTraceValid(trace))) and !trace.StartSolid or (trace.StartSolid and trace.FractionLeftSolid < 1))then
+				table.insert(origins,trace)
+				local ext = area:GetExtentInfo()
+				local bnds = Vector(ext.SizeX,ext.SizeY,ext.SizeZ + 1)
+				debugoverlay.Box( area:GetCenter() + Vector(0,0,0.5) , bnds*-0.5, bnds*0.5, 60, Color( 0, 0, 255,0 ) )
```
swampservers: remove

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,457 @@
...
+render.SetLocalModelLights( {light1} )
+render.SetAmbientLight(0.05,0.05,0.05)
+--render.SuppressEngineLighting( true )
+self:DrawModel()
+render.SuppressEngineLighting( false )
+end
+
+function MoneyPrize() -- this is where we generate a random amount of points to give the player
+	local basenumber = 100
+	for i=1,math.random(10,40) do
```
swampservers: Dude, what???

If you want a non-linear sampling probability (eg, more likely to choose a small number) do something like 

`math.pow(math.random(), 5) * (MAX-MIN) + MIN`

PYROTEKNIK: i'm not good with that kind of math lol

swampservers:
Can you remove the nav file? I'll install it

lua/entities/magicbutton.lua
swampservers: the secret is that math.random() is 0 to 1, and pow(math.random(), anynumber) is also 0 to 1. if anynumber>1, the result biases more toward 0, if anynumber < 1 the result is biased more toward 1

PYROTEKNIK: for some reason i thought pow() was the same as number^x

swampservers: pow(a,b) is a^b

PYROTEKNIK: oh right yeah i guess it makes sense that using values between 0 and 1 always returns values between 0 and 1.  usually when i use it i would use values above 1

lua/autorun/client/big_chungus_in_the_bathroom.lua
swampservers: add the header

lua/entities/kleiner/cl_init.lua
```diff
@@ -0,0 +1,154 @@
...
+	local target = self:GetTarget()
+
+	if(IsValid(target))then 
+		return target:EyePos()
+	end
+	return self:EyePos() + self:GetAngles():Forward()*600
+end
+
+
+function ENT:Think()
```
swampservers: clean up the empty function

swampservers:
@PYROTEKNIK Bro its still not formatted

PYROTEKNIK:
i ran it on that file. do you need me to format everything else?

swampservers:
Every file... why would I give you directions to batch format the entire folder if i only wanted one file?

PYROTEKNIK:
you told me you converted everything already

swampservers:
No, I formatted everything that was in contrib already. I can't format the files in your PR because they are in your fork which I can't edit.

format.bat
```diff
@@ -0,0 +1,3 @@
+@echo off
```
swampservers: I don't want this file

lua/autorun/client/big_chungus_in_the_bathroom.lua
```diff
@@ -0,0 +1,34 @@
...
+
+local function Chungus_Function(ply, key)
+    if (key == IN_USE and ply:EyePos():Distance(CHUNGUS_ENTITY:GetPos()) < 64 and ply:GetEyeTrace().HitPos:Distance(CHUNGUS_ENTITY:GetPos()) < 25) then
+        RunConsoleCommand("say_team", "lol big chungus")
+        surface.PlaySound("weapon_funnybanana/hahaha.ogg")
+    end
+end
+
+timer.Create("Chungus_Checker", 1, 0, function()
+    if (true or LocalPlayer().GetLocationName and LocalPlayer():GetLocationName() == "Bathroom") then
```
swampservers: `if true or`... ?

swampservers: GetLocationName shouldn't need to be checked for anymore, since it is in contrib

PYROTEKNIK: sorry, i can't test this stuff in cinema so i occasionally forget to remove these checks

lua/entities/kleiner/cl_init.lua
```diff
@@ -0,0 +1,139 @@
+﻿-- This file is subject to copyright - contact swampservers@gmail.com for more information.
+-- INSTALL: CINEMA
+if (SERVER) then return end
```
swampservers: this never happens

lua/entities/kleiner/cl_init.lua
```diff
@@ -0,0 +1,139 @@
...
+    end
+end
+--maybe for later? hud that counts your bounty level against kleiners
+--[[
+
+local function KleinerViolenceStarValue(star)
+	return  math.max(math.floor( (2.8^star)),1)
+end
+
+local function KleinerViolenceStars(value)
```
swampservers, line 85: dude what if it was like GTA where you kill the kleiners and then they send like armored kleiners against you lmao, and it showed 1-5 stars in the corner of the screen

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,916 @@
...
+function ENT:OnInjured(damageinfo)
+    if (self.Suicidal) then return end
+
+    if (damageinfo:GetDamage() > self:Health() / 20) then
+        local snd = table.Random(PainSounds)
+        self:Speak(snd)
+        self:AddGestureSequence(self:LookupSequence("fear_reaction_gesture"))
+    end
+
+    if (self:Health() < 0 and self.IsAlive ~= true) then
```
swampservers, line 185: `and not self.IsAlive` ?

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,916 @@
...
+    gren:PhysicsDestroy()
+    gren:SetCollisionGroup(COLLISION_GROUP_WORLD)
+    gren:SetOwner(self)
+    gren.Owner = self
+
+    return gren
+end
+
+function ENT:DropGrenades()
+    if (IsValid(self.Grenade1)) then
```
swampservers: You could do `for _,gren in ipairs({self.Grenade1,self.Grenade2}) do` (not mandatory, but that is good practice)

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,916 @@
...
+
+        timer.Simple(60 * 15, function()
+            if (IsValid(gren)) then
+                gren:Remove()
+            end
+        end)
+    end
+end
+
+if (SERVER) then
```
swampservers, line 321: unnecessary check

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,916 @@
...
+end
+
+function ENT:CanBeTarget(ent)
+    if (not IsValid(ent)) then return false end
+    if (ent == self) then return false end
+    if (ent.InVehicle and ent:InVehicle()) then return false end
+    if (ent.Alive and not ent:Alive()) then return false end
+    if (ent:GetMoveType() == MOVETYPE_FLY) then return false end
+    if (ent.IsAFK and ent:IsAFK()) then return false end
+    if (Safe and Safe(ent) and ent:IsPlayer()) then return false end
```
swampservers, line 342: Is `ent` ever not expected to be a player?

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,916 @@
...
+                if (not self.shifted) then
+                    local result = self:MoveToPos(self:GetTarget():GetPos() + VectorRand() * Vector(1, 1, 0):GetNormalized() * math.Rand(80, 150), {
+                        maxage = 50
+                    })
+
+                    if (result ~= "ok") then
+                        self:ResetBehavior()
+                    end
+
+                    --cheap method of keeping kleiners with similar targets from clumping together. i figure it's cheaper than some kind of avoidance.
```
swampservers, line 553: One thing that might work (it worked for me doing something similar in the past) is, if two kleiners intersect with each other, the one farther away from the target stops until the intersection ends

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,916 @@
...
+            self.loco:Jump()
+        end
+    end
+
+    return true
+end
+
+ENT.PathGen = function(area, fromArea, ladder, elevator, length)
+    KLPATHGEN_ITERS = (KLPATHGEN_ITERS or 0) + 1
+    local self = RELEVANT_KLEINER --this is bullshit, i guess this callback doesn't include the entity pathing.
```
swampservers, line 904: You could also make this a local function inside ENT:ChaseTarget to capture the local `self` variable

lua/entities/kleiner/sv_spawning.lua
```diff
@@ -0,0 +1,93 @@
...
+KLEINER_NPCS_FILTER = KLEINER_NPCS_FILTER or {}
+KLEINER_NPC_TARGETS = KLEINER_NPC_TARGETS or {}
+KLEINER_NPCS_CURRENT_NUMBER = KLEINER_NPCS_CURRENT_NUMBER or 0
+KLEINER_DESIRED_NUMBER = KLEINER_DESIRED_NUMBER or GetConVar("kleiner_spawncount"):GetInt() or 0
+
+function KLEINER_NPC_SETUPTIMER(value)
+    KLEINER_DESIRED_NUMBER = GetConVar("kleiner_spawncount"):GetInt() or 0
+    timer.Destroy("kleiner_spawner")
+
+    if (value > 0) then
```
swampservers: I think this check is unnecessary. Also don't see why you pass `value` to this function when it seems like its always the same value as `KLEINER_DESIRED_NUMBER` which you get inside the function anyway?

lua/entities/kleiner/sv_spawning.lua
```diff
@@ -0,0 +1,93 @@
...
+
+    --if no attacker on damage inflicted by grenade, set attacker to the grenade's owner
+    local inflictor = dmginfo:GetInflictor()
+
+    if (IsValid(inflictor) and inflictor:GetClass() == "npc_grenade_frag" and dmginfo:GetAttacker() == game.GetWorld() and IsValid(inflictor:GetOwner())) then
+        dmginfo:SetAttacker(dmginfo:GetInflictor():GetOwner())
+    end
+
+    --if grenade is hit by high force damage from non-world, change grenade owner to attacker
+    if (target:GetClass() == "npc_grenade_frag" and dmginfo:GetDamage() > 5 and dmginfo:GetDamageForce():Length() > 200 and IsValid(dmginfo:GetAttacker()) and IsValid(dmginfo:GetInflictor())) then
```
swampservers: is this so you can shoot a grenade and it explodes? uh, okay

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,463 @@
...
+ENT.AdminSpawnable = true
+ENT.PlacementSettings = {}
+ENT.PlacementSettings.RequireCeiling = false
+ENT.PlacementSettings.AllowFloor = false
+--NOTE: Generally the process will be less efficient if both of these are set to false!
+ENT.PlacementSettings.CeilingHeight = 400
+
+ENT.PlacementSettings.SurfacePropBlacklist = {"dirt", "grass", "gravel", "rock"}
+
+--**studio** is any model
```
swampservers: If models don't work, what about displacements? I think there is a special material for them too

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,463 @@
...
+            tr.mask = MASK_PLAYERSOLID
+            tr.mins = hull * Vector(1, 1, 0.1) * -0.5
+            tr.maxs = hull * Vector(1, 1, 0.1) * 0.5
+            trace = util.TraceHull(tr)
+
+            if ((not self.PlacementSettings.RequireCeiling or (trace.Hit and self:IsTraceValid(trace))) and not trace.StartSolid or (trace.StartSolid and trace.FractionLeftSolid < 1)) then
+                table.insert(origins, trace)
+                local ext = area:GetExtentInfo()
+                local bnds = Vector(ext.SizeX, ext.SizeY, ext.SizeZ + 1)
+                debugoverlay.Box(area:GetCenter() + Vector(0, 0, 0.5), bnds * -0.5, bnds * 0.5, 60, Color(0, 0, 255, 0))
```
swampservers: debug overlay?

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,463 @@
...
+        local navareas = navmesh.Find(trace2.HitPos, 128, 256, 256)
+
+        for k, v in pairs(navareas) do
+            if (v:IsUnderwater() or not v:Contains(trace2.HitPos)) then
+                navareas[k] = nil
+            end
+        end
+
+        if (table.Count(navareas) > 0) then
+            debugoverlay.SweptBox(tr2.start, trace2.HitPos, tr2.mins, tr2.maxs, Angle(), 60, Color(0, 255, 0, 0))
```
swampservers: overlay

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,463 @@
...
+
+        if (trace2 == nil) then
+            TRACE1_FAILURES = TRACE1_FAILURES + 1
+            TOTAL_FAILURES = TOTAL_FAILURES + 1
+            continue
+        end
+    end
+
+    if (trace1 and trace2) then return trace2, trace1 end --print("SUCCESS AFTER "..TOTAL_FAILURES.." ATTEMPTS!")
+    --print("SUPER BIG FAILURE! HOW IS IT POSSIBLE?")
```
swampservers: You'd probably want to use `Error()` or `assert()` in the future

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,463 @@
...
+end
+
+function ENT:Draw()
+    local pos = self:GetPos() + self:GetUp() * 3
+    local c = self:GetColor()
+    local cvector = (Vector(c.r, c.g, c.b) / 255)
+    local lc = render.ComputeLighting(pos, self:GetUp())
+    local light1 = {}
+    light1.type = MATERIAL_LIGHT_POINT
+    light1.color = cvector * 15
```
swampservers: So there is like a light on the button? what do you think about animating the intensity over time with `sin(SysTime())` ?

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,463 @@
...
+    --render.SuppressEngineLighting( true )
+    self:DrawModel()
+    render.SuppressEngineLighting(false)
+end
+
+-- this is where we generate a random amount of points to give the player
+function MoneyPrize()
+    local basenumber = 100
+
+    for i = 1, math.random(10, 40) do
```
swampservers: can you replace this with actual math please

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,463 @@
...
+        basenumber = (basenumber) ^ math.Rand(1.01, 1.06)
+    end
+
+    basenumber = math.Clamp(basenumber, 1000, math.random(100000, 500000))
+    basenumber = math.Round(basenumber, -3)
+
+    return basenumber * 1
```
swampservers: *1

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,463 @@
...
+}
+
+--NOTE: On these functions, please return a string if there was a success, and nil if there was not.
+function ENT:Effect(ply)
+    local effect
+    local item
+    local counter = 0
+
+    while (item == nil and counter < 100) do
+        effect = table.Copy(table.Random(MagicButtonOutcomes))
```
swampservers: Don't see why you need to copy it

lua/entities/magicbutton.lua
```diff
@@ -0,0 +1,463 @@
...
+    },
+}
+
+--NOTE: On these functions, please return a string if there was a success, and nil if there was not.
+function ENT:Effect(ply)
+    local effect
+    local item
+    local counter = 0
+
+    while (item == nil and counter < 100) do
```
swampservers: Maybe you can shuffle the order of `MagicButtonOutcomes`, then iterate it, and return early if one of the effects "works"?

lua/autorun/client/big_chungus_in_the_bathroom.lua
```diff
@@ -0,0 +1,38 @@
+﻿-- This file is subject to copyright - contact swampservers@gmail.com for more information.
+-- INSTALL: CINEMA
+BIG_CHUNGUS_LOCATION = Vector(0, 1598, 64)
+BIG_CHUNGUS_NORMAL = Vector(1, 0, 0)
+if (game.GetMap() ~= "cinema_swamp_v3") then return end
+if (engine.ActiveGamemode() ~= "cinema") then return end
```
swampservers: remove check

swampservers:
This is looking good, I'll test it soon and see if there are any other issues!

lua/autorun/client/big_chungus_in_the_bathroom.lua
swampservers: Shouldn't need the checks anymore for location-related stuff

lua/entities/kleiner/init.lua
PYROTEKNIK: wouldn't that require extra server work doing collisions with eachother

lua/entities/kleiner/cl_init.lua
PYROTEKNIK: i could do something like that if you're interested. I just wasn't sure if i wanted to include the current hud since it looked kind of shitty. I'm not sure about armor since kleiners are meant to interact with other players too. I can look into this if you'd like, though.

lua/entities/kleiner/sv_spawning.lua
PYROTEKNIK: value is passed from the convar's change callback to update it.  i should change these to actually use the passed value

lua/entities/magicbutton.lua
PYROTEKNIK: some displacements are not terrain. These checks are intended to prevent placing them on surfaces that don't make sense for a button to be, like floating on a disjointed part of a model's collision geometry, or on natural surfaces.

lua/entities/magicbutton.lua
PYROTEKNIK: I swear i did that already. I think i lost that change.

lua/entities/magicbutton.lua
PYROTEKNIK: These are debugging functions used for visualizing. They apparently do nothing unless developer is set to 1.

lua/entities/kleiner/sv_spawning.lua
```diff
@@ -0,0 +1,93 @@
...
+    KLEINER_DESIRED_NUMBER = value
+    timer.Destroy("kleiner_spawner")
+
+    if (KLEINER_DESIRED_NUMBER > 0) then
+        timer.Create("kleiner_spawner", 1, 0, function()
+            if (table.Count(KLEINER_NPCS) < KLEINER_DESIRED_NUMBER) then
+                local newkleiner = ents.Create("kleiner")
+
+                if (IsValid(newkleiner)) then
+                    local spawnpoint = gmod.GetGamemode().SpawnPoints and table.Random(gmod.GetGamemode().SpawnPoints):GetPos() or Vector(0, 0, 16)
```
swampservers: This line threw a NULL entity error

swampservers: probably just ents.FindByClass("info_player_start") then cache it, and if you find a NULL entity, update the cache

lua/entities/kleiner/sv_spawning.lua
```diff
@@ -0,0 +1,93 @@
+﻿-- This file is subject to copyright - contact swampservers@gmail.com for more information.
+-- INSTALL: CINEMA
+CreateConVar("kleiner_spawncount", "18", FCVAR_ARCHIVE + FCVAR_PROTECTED, "Number of Kleiner NPCs to spawn automatically.", 0, 100)
+KLEINER_NPCS = KLEINER_NPCS or {}
+KLEINER_NPCS_FILTER = KLEINER_NPCS_FILTER or {}
+KLEINER_NPC_TARGETS = KLEINER_NPC_TARGETS or {}
+KLEINER_NPCS_CURRENT_NUMBER = KLEINER_NPCS_CURRENT_NUMBER or 0
+KLEINER_DESIRED_NUMBER = KLEINER_DESIRED_NUMBER or GetConVar("kleiner_spawncount"):GetInt() or 0
+
+function KLEINER_NPC_SETUPTIMER(value)
```
swampservers: this is overcomplicated and doesn't work when i change the convar. Just do `timer.Create` in the global scope (eg. not inside a function) and check the `GetConVar("kleiner_spawncount"):GetInt()` every time. There is no reason to use `hook.Add("Initialize"`; just put it in the global scope.

lua/entities/kleiner/init.lua
```diff
@@ -0,0 +1,911 @@
...
+            if (math.random(1, 50) == 1) then
+                self:ResetBehavior()
+            else
+                local wanderpos
+
+                if (self.WanderForcePos) then
+                    wanderpos = self.WanderForcePos + VectorRand() * Vector(1, 1, 0):GetNormalized() * math.Rand(80, 150)
+                end
+
+                wanderpos = wanderpos or self:FindSpot("random", {
```
swampservers: This seems to be the main lag causer, looking at https://github.com/Facepunch/garrysmod/blob/be251723824643351cb88a969818425d1ddf42b3/garrysmod/gamemodes/base/entities/entities/base_nextbot/sv_nextbot.lua it seems to essentially find a bunch of nearby areas, then make a path to each area (!!!). We should probalby use a much lower radius so it finds less areas, and it can find its way to distant areas after multiple wanders. It might also work to override ent:FindSpots to not calculate a path and just use the vector:Distance() instead.

lua/autorun/swamp_core.lua
```diff
@@ -57,7 +57,9 @@ function math.power2(n)
     return math.pow(2, math.ceil(math.log(n) / math.log(2)))
 end
 
--- TABLE
+BLACK = Color(0, 0, 0, 255)
```
swampservers, line 60: there shouldn't be my changes in this PR, i think you need to fetch and merge upstream/master then push to origin/master

PYROTEKNIK:
This is where I had originally inserted vec:inbox

PR #225, opened 2021-04-04
Add reload sound for weapon_squee

ghost:
I added a reload sound for weapon_squee: "mowsquee.wav"

This is a recreation of a popular sound emote on the server combining mow.wav and squee.wav. Code for the sound emote: `{2,0,0,0.2,mow:2,0.2,0.2,4,squee}`

Let me know if I should change it to an alternative sound such as squee2.ogg or squee3.ogg.


lua/weapons/weapon_squee/shared.lua
```diff
@@ -15,3 +15,7 @@ end
 function SWEP:SecondaryAttack()
 	self:ExtEmitSound("boop.wav", {shared=true})
 end
+
+function SWEP:Reload()
```
swampservers, line 19: I'm pretty sure :Reload() fires as full auto while its held down, so this needs some extra logic (look at another soundgun with a reload sound)

ghost: Yep, forgot to add that. Added

swampservers: Delay is too short

ghost: I increased it. Now it has time for the full sound to play plus a small buffer

ghost:
@swampservers Don't think the sound has been added fully. Code works on the server but the sound doesn't play.

PR #233, opened 2021-04-29
New pointshop accessory

Aryanne-Hoofler:
added a new hat to pointshop


swampservers:
Looks correct, thanks!

PR #236, opened 2021-05-01
File Fix

Noztik:
your network or whatever is causing issues and it's leading to this check being disrupted so I just have to disable it


swampservers:
BTW https://github.com/timschumi/gmod-chttp/issues/16
so i should be able to fix any filtering related issues now

PR #237, opened 2021-05-03
Theater Stuff

Noztik:


gamemodes/cinema/gamemode/modules/scoreboard/cl_settings.lua
```diff
@@ -174,6 +174,8 @@ function SETTINGS:Create()
     --addCheckbox('Don\'t load chat images',"fedorachat_hideimg","Hides all images in chat")
     addCheckbox('Hide interface', "cinema_hideinterface", "Clean Your Screen")
     addCheckbox('Hide players in theater', "cinema_hideplayers", 'For when trolls stand in front of your screen')
+    addCheckbox('Trust Videos', "cinema_trust_videos", 'Don\'t show the warning on untrusted videos')
```
swampservers: This setting is probably just going to be confusing. How about instead, it shows "run cinema_load_all_urls 1 to never see this again" on the warning overlay? Its in cl_draw.lua

gamemodes/cinema/gamemode/modules/theater/sh_video.lua
```diff
@@ -206,15 +206,20 @@ if SERVER then
...
 
     --used by services
     function TRUSTED_VIDEO_HOST(k)
-        k = k:lower():gsub("https", "http"):gsub("http://", ""):gsub("www.", "")
+        local t = url.parse2(k)
+        if not t.host then return false end
+        CURRENTKEY = t.host
 
-        return k:StartWith("swampservers.net/") or k:StartWith("dropbox.com/") or k:StartWith("puu.sh/") or k:StartWith("discord.gg/") or k:StartWith("discord.com/")
+        return t.host == "www.swampservers.net" or t.host == "www.dropbox.com" or t.host == "www.puu.sh" or t.host == "cdn.discordapp.com"
```
swampservers: Wait so it puts www on everything, even if the url doesn't have www?

gamemodes/cinema/gamemode/modules/theater/sh_video.lua
```diff
@@ -229,6 +234,7 @@ if CLIENT then
 
                     if v and not v:ShouldTrust() then
                         v.FORCETRUST = true
+                        if CURRENTKEY then TEMP_TRUST[CURRENTKEY] = true end
```
swampservers: this is dumb, just url.parse2 the v:Key() and get the host again, assuming you're trying to make a set of trusted hosts

gamemodes/cinema/gamemode/modules/theater/sh_video.lua
Noztik: the host member has the www in the string when I tested it so it's www unless there's a special case like cdn.discordapp.com

gamemodes/cinema/gamemode/modules/theater/sh_video.lua
Noztik: I forgot I could do that to video objects

gamemodes/cinema/gamemode/modules/theater/sh_video.lua
```diff
@@ -206,15 +206,18 @@ if SERVER then
 end
 
 if CLIENT then
+	local TrustConvar = CreateClientConVar("cinema_load_all_urls","0",true,false,"",0,1)
+    local TEMP_TRUST = {}
```
swampservers: I think we should remove the TRUSTED_VIDEO_HOST function below, and just change it to a table, which contains the urls like dropbox by default, and when F8 is pressed, the current host gets added to that same table. Then update the services to look in the TRUSTED_VIDEO_HOST table rather than calling the TRUSTED_VIDEO_HOST function.

swampservers:
Let me suggest a cleaner structure:
1. Remove SERVICE:ShouldTrust (anywhere it appears)
2. Add SERVICE:GetHost(video) which by default (in sh_base) returns nil, but for file-type services it gets the url.host out of the data
3. Make VIDEO:ShouldTrust like:
```
function VIDEO:ShouldTrust()
    local host = self:Service():GetHost(self)
    return TrustConvar:GetInt()==1 or host==nil or TRUSTED_HOST[host]
end
    ```

swampservers:
Looks great!

PR #250, opened 2021-05-24
Spraypaint exploit fix, better error handling

PYROTEKNIK:
no longer flips out and breaks everything if shit is missing.
should crash significantly less often
users can no longer set the decal to anything they want through console


PYROTEKNIK:
This pr has a lot more than i expected because when i went into singleplayer to fix exploits it suddenly started having serious crashing issues. i don't know if the previous branch will affect the server in the same way.  this branch should hopefully address an overall lack of stability. most of the crashes seem to occur when accessing materials that aren't loading properly

PR #254, opened 2021-06-14
Swamp devtools

Phoenix7302:
Adds devtools for client-side file editing.
Commands are dev, setroot, cleardev, and frefresh. Files should refresh recursively every 2 seconds.
Let me know on code feedback.

### DIFF
--- a/lua/autorun/client/cl_devtools.lua
+++ b/lua/autorun/client/cl_devtools.lua
@@ -1,133 +1,113 @@
-local ranOnce = !ranOnce
-if (!ranOnce) then return end -- just to prevent edge cases
-ranOnce = true
+-- This file is subject to copyright - contact swampservers@gmail.com for more information.
+-- INSTALL: CINEMA
 
-local isDev = isDev
-local refreshTime = 2
-local rootFolder = rootFolder or nil
-local fileTimes = fileTimes or {}
-local devReplStr = nil -- classname of all child files, taken from folder
-
-local function replaceFunction(base, regex, replacement)
-	local iStart, iEnd = string.find(base, regex, 1)
-
-	while iStart do
-		local xStart, _ = string.find(base, "(", iEnd, true) -- func name index
-		base = string.sub(base, 1, iStart-1) ..
-			replacement .. "." .. string.sub(base, iEnd + 1, xStart-1) ..
-			" = function" .. string.sub(base, xStart)
-		iStart, iEnd = string.find(base, regex, xStart)
+SWAMP_DEV = SWAMP_DEV or {}
+SWAMP_DEV.isDev = SWAMP_DEV.isDev
+SWAMP_DEV.refreshDelay = 2
+SWAMP_DEV.itemStruct = SWAMP_DEV.itemStruct
+SWAMP_DEV.fileTimes = SWAMP_DEV.fileTimes or {}
+SWAMP_DEV.structTypes = {
+	["weapons"] = function(curClass) SWEP = weapons.GetStored(curClass) end,
+	["entities"] = function(curClass) ENT = scripted_ents.GetStored(curClass) end,
+	["effects"] = function(curClass)
+		local fxTab = effects.GetList()
+		local fx = "effects/" .. curClass
+		for _, v in ipairs(fxTab) do
+			if (v.Folder == fx) then
+				EFFECT = v
+				return
+			end
+		end
 	end
-	return base
-end
+}
 
-local function refreshFile(path)
-	if (!file.Exists(path, "LUA")) then return print("File " .. path .. " doesn't exist") end
-	if (!isDev) then return end
-	local code = file.Read(path, "LUA")
-
+-- Returns a string to print
+local function refreshFile(path, filename)
+	local code = file.Read(path, "MOD")
 	-- current filename as class format
-	local curClass = devReplStr or string.TrimRight(string.gsub(path, "^.*/", ""), ".lua")
+	if SWAMP_DEV.itemStruct then
+		local parentFolder = string.gsub(string.sub(path, 1, -#filename - 2), ".*/", "")
+		-- if parent isn't a struct then use parent as class, else use filename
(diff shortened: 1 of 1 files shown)


lua/autorun/client/cl_devtools.lua
```diff
@@ -0,0 +1,133 @@
+local ranOnce = !ranOnce
```
swampservers: Don't put this in. The code should work if you rerun the file.

lua/autorun/client/cl_devtools.lua
```diff
@@ -0,0 +1,133 @@
...
+if (!ranOnce) then return end -- just to prevent edge cases
+ranOnce = true
+
+local isDev = isDev
+local refreshTime = 2
+local rootFolder = rootFolder or nil
+local fileTimes = fileTimes or {}
```
swampservers: I'm pretty sure what this is trying to do is read the global fileTimes and store it into a different local variable. If the file should be re-runnable without overriding the value of filetimes, make it a global

lua/autorun/client/cl_devtools.lua
```diff
@@ -0,0 +1,133 @@
...
+local devReplStr = nil -- classname of all child files, taken from folder
+
+local function replaceFunction(base, regex, replacement)
```
swampservers: Isn't this basically just gsub

lua/autorun/client/cl_devtools.lua
```diff
@@ -0,0 +1,133 @@
...
+	-- performance impact should be minimal, it's ran once for testing, anyway
+	local storedStr = ".GetStored(\"" .. curClass .. "\")"
+
+	-- replace variables
+	code = string.gsub(string.gsub(code,
+		"%f[%a?]SWEP%.", "weapons" .. storedStr .. "."),
+		"%f[%a?]ENT%.", "scripted_ents" .. storedStr .. ".")
+
+	-- set all function definitions to assign to table
+	code = replaceFunction(replaceFunction(code,
```
swampservers: I think it would be easier to just set the global SWEP = scripted_weapons.GetStored(), then run the file. However, you do need to figure out how to handle `include` - either overwrite it globally or gsub it in the code string

lua/autorun/client/cl_devtools.lua
```diff
@@ -0,0 +1,133 @@
...
+		if (!isDev or rootFolder != oldFolder) then
+			return print("Stopped automatic refresh for " .. oldFolder)
+		end
+		timedRefresh()
+	end)
+end
+
+
+-- enable dev editing
+concommand.Add("dev",
```
swampservers, line 85: You have three commands here - `dev`, `setroot`, `cleardev`, but you only need one. Just make it be `dev` to toggle editing and have the root always be addons/contrib . It is hard to remember commands like that.

lua/autorun/client/cl_devtools.lua
```diff
@@ -0,0 +1,133 @@
...
+		if ((rootFolder == "weapons/") or (rootFolder == "entities/")) then
+			devReplStr = v
+		end
+		times[v] = recurseRefresh(times[v], v .. "/")
+	end
+	devReplStr = nil
+	return times
+end
+
+local function timedRefresh()
```
swampservers: you should probably use timer.Create, and have the timer just always be running but return early if !isDev . That is a much more stable way to code things as there is no possibility of multiple versions of the timer running or whatever

lua/autorun/client/cl_devtools.lua
```diff
@@ -0,0 +1,133 @@
...
+	function()
+		isDev = false
+		fileTimes = {}
+		print("Dev cleared")
+	end,
+	nil, "Stop checking files, exit dev mode", FCVAR_UNREGISTERED)
+
+
+-- force refresh file
+concommand.Add("frefresh", function(_, _, args)
```
swampservers: dev_refresh so it appears in autocomplete for dev

lua/autorun/client/cl_devtools.lua
```diff
@@ -0,0 +1,133 @@
...
+			replacement .. "." .. string.sub(base, iEnd + 1, xStart-1) ..
+			" = function" .. string.sub(base, xStart)
+		iStart, iEnd = string.find(base, regex, xStart)
+	end
+	return base
+end
+
+local function refreshFile(path)
+	if (!file.Exists(path, "LUA")) then return print("File " .. path .. " doesn't exist") end
+	if (!isDev) then return end
```
swampservers: Why check this? Structure your code so you know it's not being called in the first place.

lua/autorun/client/cl_devtools.lua
```diff
@@ -0,0 +1,133 @@
...
+
+	RunString(code, "swampDevTools")
+	print("File refreshed: " .. path)
+end
+
+local function recurseRefresh(times, curPath)
+	local files, folders = file.Find(rootFolder .. curPath .. "*", "LUA")
+	for k, v in pairs(files) do
+		local newtime = file.Time(rootFolder .. curPath .. v, "LUA")
+		if (!times[k]) then -- new file created, or first run
```
swampservers:
```
if (times[k] or newtime) ~= newtime then
    refreshFile(rootFolder .. curPath .. v)
end
times[k] = newtime	
```

swampservers:
Thank you for your fast turnaround! It looks like you have the basic idea correct but the code is more complicated than necessary and I suspect some functionality issues. I see unnecessary checks put in which is a big red flag because it seems like you're not really sure when certain functions will be called. Putting extra checks in "to be safe" is generally bad practice; good practice is writing code that provably won't ever trigger those conditions in the first place.

I think this won't work because you're using the "LUA" file.Read mode which IIRC is all mounted Lua. On the server, the addon isn't going to be mounted, it's basically sideloaded.

Phoenix7302:
Really helpful feedback, thanks! This commit should work on servers.

swampservers:
Thanks, i'll try it out. Email me with how you'd like to receive payment.

PR #255, opened 2021-06-16
refactor: developer tools.

NoSharp:
This is a near complete refactor of how the previous iteration of this developer tools utility that was made.

**Please test this properly before deploying to production. I cannot test this on my single player game.**

### DIFF
--- a/lua/autorun/client/cl_devtools.lua
+++ b/lua/autorun/client/cl_devtools.lua
@@ -12,7 +12,7 @@
 SWAMP_DEV = SWAMP_DEV or {}
 
 SWAMP_DEV.allowed = {
-    ["STEAM_0:0:105777797"] = true
+    ["SteamID Here."] = true
 }
 
 -- Are we in development mode?
@@ -59,22 +59,23 @@ do
     local loadFunc = function() end
     function loadFile(filePath, environmentVars)
         local code = file.Read(filePath, "MOD")
-        if not code then error("No file exists with the file path: " .. filePath) end
+        if not code then 
+            error("No file exists with the file path: " .. filePath)
+        end
         environmentVars = environmentVars or {}
-        loadFunc = CompileString(code, "SwampDevTools/" .. filePath)
+       
         -- setup our environment
         local environment = _G
         for key,var in pairs(environmentVars) do
             environment[key] = var
         end
-        
-        debug.setfenv(loadFunc,environment)
-
-        xpcall(loadFunc, function(err)
+        xpcall(function()
+            RunString(code, "SwampDevTools/" .. filePath)
+            
+        end, function(err)
             print("[contrib] Error during loading file with devtools:", err)
             debug.Trace()
         end)
-
         -- remove the variaables from the global.
         for key,_ in pairs(environmentVars) do
             environment[key] = nil
@@ -128,7 +129,7 @@ concommand.Add("dev", function()
 
     SWAMP_DEV.enabled = not SWAMP_DEV.enabled
 
-    if  SWAMP_DEV.enabled then
+    if SWAMP_DEV.enabled then
         print("Dev mode enabled")
 
         timer.Create("SWAMP_DEV.Refresh", SWAMP_DEV.refreshDelay, 0, function()
@@ -148,17 +149,23 @@ concommand.Add("dev_refresh", function(_, _, args)
     local filePath = args[1]
     if not SWAMP_DEV.enabled or not filePath then return end
     print("Attempting to force refresh file " ..filePath)
+    
     local structFound 
+    local ran
     for fileType in filePath:gmatch("[^/\\]+") do
         
         if structFound then
             loadFile(filePath, structFound(fileType)) 
+            ran = true
             break
         end
 
         structFound = SWAMP_DEV.structEnvironments[fileType]
     end
-    
-    
+
+    if not ran then 
+        loadFile(filePath, {})
+    end
+
     print("File " .. filePath:gsub(".*/", "") .. " force refreshed")
 end, nil, "Force refresh file PATH, CLASSNAME (if ENT or SWEP)", FCVAR_UNREGISTERED)
--- a/lua/autorun/client/cl_devtools.lua
+++ b/lua/autorun/client/cl_devtools.lua
@@ -1,4 +1,13 @@
-﻿-- This file is subject to copyright - contact swampservers@gmail.com for more information.
+﻿--[[
+    Limitations:
+        - If a path has lua/STRUCTHERE eg. "lua/autorun/lua/weapons/test.lua" then
+            It'll pass the Struct into it, not sure if it's a bug or a feature.
+        - A file within a sub directory of a struct folder like: "lua/weapons/testEnt/mysubdir/lua"
+            Won't have the SWEP/struct table passed to it.
+
+]]--
+
+-- This file is subject to copyright - contact swampservers@gmail.com for more information.
 -- INSTALL: CINEMA
 SWAMP_DEV = SWAMP_DEV or {}
 -- Are we in development mode?
@@ -9,21 +18,25 @@ SWAMP_DEV.refreshDelay = 2
 local currentItemStruct = nil
 
 --- This stores file times for certain root folder.
---- Each sub table will have a Table of FileNames[key] -> SaveTime[value]
-SWAMP_DEV.fileTimes = SWAMP_DEV.fileTimes or {
-    ["lua"] = {},
-    ["gamemodes"] = {}
-}
+--- FileNames[key] -> SaveTime[value]
+SWAMP_DEV.fileTimes = SWAMP_DEV.fileTimes or {}
+ 
+-- lua/STRUCT/*
 
 SWAMP_DEV.structEnvironments = {
     ["weapons"] = function(curClass)
+        local found = weapons.GetStored(curClass)
+        if not found then return end
         return {
-            SWEP = weapons.GetStored(curClass)
+            SWEP = found
         }
     end,
     ["entities"] = function(curClass)
+        local found = scripted_ents.GetStored(curClass)
+        if not found then return end
+       
         return {
-            ENT = scripted_ents.GetStored(curClass)
+            ENT = found
         }
     end,
     ["effects"] = function(curClass)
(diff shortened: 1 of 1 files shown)


lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
+            Won't have the SWEP/struct table passed to it.
+
+]]--
+
+-- This file is subject to copyright - contact swampservers@gmail.com for more information.
 -- INSTALL: CINEMA
 SWAMP_DEV = SWAMP_DEV or {}
-SWAMP_DEV.isDev = SWAMP_DEV.isDev
+-- Are we in development mode?
+SWAMP_DEV.isDev = false
```
TomDotBat: Could just do a nil check, no real need for the assignment to false here.

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
+local rootDirectory = "addons/contrib/"
+local refreshFolder
+local loadFile
+do 
+    local loadFunc = function() end
+    function loadFile(filePath, environmentVars)
+        local code = file.Read(filePath, "MOD")
+        if not code then error("No file exists with the file path: " .. filePath) end
+        environmentVars = environmentVars or {}
+        loadFunc = CompileString(code, "SwapDevTools/" .. filePath)
```
TomDotBat: Swamp*

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
+        -- setup our environment
+        local environment = _G
+        for key,var in pairs(environmentVars) do
```
TomDotBat, line 68: Would be nice to keep spacing consistency between args/for loop vars.

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
+            environment[key] = var
         end
-    end
+        
+        debug.setfenv(loadFunc,environment)
+
+        xpcall(loadFunc, function(err)
+            print("[contrib] Error during loading file with devtools:", err)
```
TomDotBat, line 75: Potentially not wrong - but check if this follows the common logging format in this project.

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
+            refreshFolder(subDir .. "/" .. folder)
         end
     end
-
-    return times
 end
 
 -- toggle dev editing. Root is addons/contrib
 concommand.Add("dev", function()
     if (LocalPlayer():GetRank() <= 0) then return end
```
TomDotBat: Confirm that this is actually not going to open up dev tools to donators or something.

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
-    -- current filename as class format
-    if SWAMP_DEV.itemStruct then
-        local parentFolder = string.gsub(string.sub(path, 1, -#filename - 2), ".*/", "")
 
-        -- if parent isn't a struct then use parent as class, else use filename
-        if (not SWAMP_DEV.structTypes[parentFolder]) then
-            SWAMP_DEV.structTypes[SWAMP_DEV.itemStruct](parentFolder)
-        else
-            SWAMP_DEV.structTypes[SWAMP_DEV.itemStruct](string.TrimRight(string.gsub(path, "^.*/", ""), ".lua"))
+local rootDirectory = "addons/contrib/"
```
TomDotBat: As this is constant consider using UPPER_SNAKE_CASE.

TomDotBat (changes requested):
Looks alright to me - just a few small changes and verification needed from @swampservers.

lua/autorun/client/cl_devtools.lua
swampservers: We automatically glualint all the files, so it's not really necessary to enforce perfect formatting

lua/autorun/client/cl_devtools.lua
swampservers: Rank is only for staff. However I was actually thinking we should have a hardcoded steamID list in here so devs don't have to be staff to use these tools. Then i'd update the readme to say "Make a PR to add your name to the developer list" which is cool because if someone is incapable of that they are obviously not worth spending any time helping.

lua/autorun/client/cl_devtools.lua
swampservers: There's not really a standard

TomDotBat: > There's not really a standard

Would you say this is fine as is then?

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
 SWAMP_DEV.fileTimes = SWAMP_DEV.fileTimes or {}
+ 
+-- lua/STRUCT/*
 
-SWAMP_DEV.structTypes = {
+SWAMP_DEV.structEnvironments = {
     ["weapons"] = function(curClass)
-        SWEP = weapons.GetStored(curClass)
+        local found = weapons.GetStored(curClass)
+        if not found then return end
```
swampservers: I'd prefer like `found and {SWEP=found} or nil`

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
 
-    -- only replace includes that don't use variables
-    RunString(string.gsub(code, "include%s*%(%s*[\'\"].-[\'\"]%s*%)", ""), "swampDevTools")
-
-    return filename
-end
-
-local function recurseRefresh(root, times, curPath)
-    local files, folders = file.Find(root .. curPath .. "*", "MOD")
+    local fileTime = file.Time
```
swampservers, line 85: What's the point of this?

lua/autorun/client/cl_devtools.lua
swampservers: "enabled" would make more sense to name this

lua/autorun/client/cl_devtools.lua
NoSharp: Localization reduces indexing times, as this is called frequently this will be slightly more efficient.

lua/autorun/client/cl_devtools.lua
NoSharp: Addressed in a more recent comment.

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
+        environmentVars = environmentVars or {}
+        loadFunc = CompileString(code, "SwapDevTools/" .. filePath)
+        -- setup our environment
+        local environment = _G
+        for key,var in pairs(environmentVars) do
+            environment[key] = var
         end
-    end
+        
+        debug.setfenv(loadFunc,environment)
```
swampservers: Aren't you just setting _G? I don't see why this is necessary

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
+local loadFile
+do 
+    local loadFunc = function() end
+    function loadFile(filePath, environmentVars)
+        local code = file.Read(filePath, "MOD")
+        if not code then error("No file exists with the file path: " .. filePath) end
+        environmentVars = environmentVars or {}
+        loadFunc = CompileString(code, "SwapDevTools/" .. filePath)
+        -- setup our environment
+        local environment = _G
```
swampservers: Why not just set members of _G?

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,165 @@
...
-            SWAMP_DEV.structTypes[SWAMP_DEV.itemStruct](string.TrimRight(string.gsub(path, "^.*/", ""), ".lua"))
+local rootDirectory = "addons/contrib/"
+local refreshFolder
+local loadFile
+do 
+    local loadFunc = function() end
+    function loadFile(filePath, environmentVars)
+        local code = file.Read(filePath, "MOD")
+        if not code then error("No file exists with the file path: " .. filePath) end
+        environmentVars = environmentVars or {}
```
swampservers, line 66: delete this line and do `for k,v in pairs(environmentVars or {})`

lua/autorun/client/cl_devtools.lua
NoSharp: No, it's not just setting _G, as we're loading the environment of the file, the idea is that in the future this could be used to sandbox the code being reloaded to an extent.

lua/autorun/client/cl_devtools.lua
NoSharp: Addressed in the above comment

lua/autorun/client/cl_devtools.lua
NoSharp: as it's *currently* being iterated over twice, it'd be more optimal to the comparison once than twice in each loop.

lua/autorun/client/cl_devtools.lua
TomDotBat: Maybe it'd be better to put this table in a different location and made global so we can reference it in other systems. Perhaps if you wanted a contributor tag on player overheads or something, etc.

swampservers:
> This is a near complete refactor of how the previous iteration of this developer tools utility that was made.
> 
> **Please test this properly before deploying to production. I cannot test this on my single player game.**

You should be able to test it; start up a singleplayer game without contrib installed, and with this script in garrysmod/lua/autorun, then put the contrib folder in place while the game is running?

lua/autorun/client/cl_devtools.lua
swampservers: ah I see. The minute performance difference is totally irrelevant, this isn't a critical section of code and it won't even be running on most clients anyway. But putting that check logic in twice would be redundant.

lua/autorun/client/cl_devtools.lua
swampservers: How is that potentially useful? The point of this code is the ability to test clientside changes to contrib on the live server. The committed code doesn't run with any special environment, so, why would you want to do that?

NoSharp:
That was more in regards to the GetRank check, which is now obsolete.

lua/autorun/client/cl_devtools.lua
NoSharp: It was for a potential feature. I can remove it, if you'd like that's fine.

swampservers: Yeah i would say remove it, I am curious what the usefulness of that is though

NoSharp:
That should bring it up to what you'd like, however you'll probably need to fill in steamid's.

lua/autorun/client/cl_devtools.lua
NoSharp: The intended feature of that is so that you could add a certain layer of sandbox-ing around reloaded scripts, so if you didn't want certain reloaded scripts having access to certain global functions that's the place to do it. This was prior to my understanding of you wanting a hard coded list of allowed users.

NoSharp:
bump: anymore on this?

swampservers:
Oh sorry I totally missed the notification that this was updated!

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,171 @@
...
+
+--- This stores file times for certain root folder.
+--- FileNames[key] -> SaveTime[value]
 SWAMP_DEV.fileTimes = SWAMP_DEV.fileTimes or {}
+ 
+-- lua/STRUCT/*
 
-SWAMP_DEV.structTypes = {
+SWAMP_DEV.structEnvironments = {
     ["weapons"] = function(curClass)
```
swampservers, line 32: I just realized this is missing gamemode (gamemode files use GM, not the same as GAMEMODE)

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,171 @@
...
+﻿--[[
+    Limitations:
+        - If a path has lua/STRUCTHERE eg. "lua/autorun/lua/weapons/test.lua" then
+            It'll pass the Struct into it, not sure if it's a bug or a feature.
+        - A file within a sub directory of a struct folder like: "lua/weapons/testEnt/mysubdir/lua"
+            Won't have the SWEP/struct table passed to it.
+
+]]--
+
+-- This file is subject to copyright - contact swampservers@gmail.com for more information.
```
swampservers: this should be first line

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,171 @@
...
 -- INSTALL: CINEMA
 SWAMP_DEV = SWAMP_DEV or {}
-SWAMP_DEV.isDev = SWAMP_DEV.isDev
+
+SWAMP_DEV.allowed = {
+    ["SteamID Here."] = true
```
swampservers: put your own steamid(s) with comment with your nickname

lua/autorun/client/cl_devtools.lua
swampservers: I really don't think the performance impact will matter. Code is cleaner without it.

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,171 @@
...
-        local parentFolder = string.gsub(string.sub(path, 1, -#filename - 2), ".*/", "")
 
-        -- if parent isn't a struct then use parent as class, else use filename
-        if (not SWAMP_DEV.structTypes[parentFolder]) then
-            SWAMP_DEV.structTypes[SWAMP_DEV.itemStruct](parentFolder)
-        else
-            SWAMP_DEV.structTypes[SWAMP_DEV.itemStruct](string.TrimRight(string.gsub(path, "^.*/", ""), ".lua"))
+local ROOT_DIRECTORY = "addons/contrib/"
+local refreshFolder
+local loadFile
```
swampservers, line 56: what's the point of putting these here and then making a `do ... end`, it's messy

lua/autorun/client/cl_devtools.lua
swampservers: You can leave it in if you want, but I think the overwhelming majority of the lag from this function will be due to the file.Time call looking at the disk synchronously.

If you're really worried about performance, what you could do is keep track of the 5 most recently edited files and update them 3x per second, and then update other files every 5 seconds, and additionally, don't do them all at once, spread them out over the 5 seconds and do a few every frame.

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,171 @@
...
-            print("INCLUDING: " .. refreshFile(includePath, string.gsub(reqInclude, ".*/", "")))
-            -- Then check the lua/ game folder
-        elseif (file.Exists("addons/contrib/lua/" .. reqInclude, "MOD")) then
-            print("INCLUDING: " .. refreshFile("addons/contrib/lua/" .. reqInclude, string.gsub(reqInclude, ".*/", "")))
-        else
-            print("Include handled by file")
+        environmentVars = environmentVars or {}
+       
+        -- setup our environment
+        local environment = _G
```
swampservers: clean this up?

lua/autorun/client/cl_devtools.lua
```diff
@@ -1,144 +1,171 @@
...
-    for s in string.gmatch(args[1], "[^/\\]+") do
-        if SWAMP_DEV.structTypes[s] then
-            SWAMP_DEV.itemStruct = s
+    local filePath = args[1]
+    if not SWAMP_DEV.enabled or not filePath then return end
+    print("Attempting to force refresh file " ..filePath)
+    
+    local structFound 
+    local ran
+    for fileType in filePath:gmatch("[^/\\]+") do
```
swampservers, line 155: Is there logic here that can be shared with the other refresh function? Maybe just have a flat list of all files, then whenever one is updated, do like RunConsoleCommand("dev_refresh", filename)

swampservers:
I'll pay you $60 instead since I missed the notification

NoSharp:
alright thanks

lua/autorun/client/cl_devtools.lua
NoSharp: it minimises the scope that gets used and is an optimisation.

lua/autorun/client/cl_devtools.lua
NoSharp: How so?

lua/autorun/client/cl_devtools.lua
NoSharp: Not really as this traverses the path, whereas, in the the automatic refresh it's loading each folder sepeartely.

lua/autorun/client/cl_devtools.lua
swampservers: why are you changing the name of this variable instead of just using _G itself? and don't say "for optimization", even if indexing the local variable is 1ns faster that is totally irrelevant in the context of this function which is not critical at all

NoSharp: Its for performance as is everything i do. I focus on optimising everything as much as possible no mattter its calling context.

swampservers: Okay, but, this kind of "optimization" is totally missing the forest because of the trees.

First of all this isn't in a PrePlayerDraw hook. This code only runs _when a developer is using the devtools_ and _they have just edited a file_. That happens probably less than 100 times per day on average, and happens only when the developer is editing something, so we really don't need to give them the extra 0.00001% frames per second this kind of edit would realize (if you generously estimate that this saves 100ns and say they are updating one file every second). Clearly this doesn't make any significant difference, especially compared to something like PrePlayerDraw, which I estimate probably gets run a cumulative 20 billion times per day by our players.

At the same time though this addon _does_ have a significant performance issue in it. The problem is how when checking the file modification times, you check all of the files at once every two seconds rather than amortizing those checks over the two seconds. This is going to spike the frame time every time you check it. I wouldn't be surprised if the spike is quite significant as well, because I assume file.Time requires a synchronous disk read which might take something like 1ms (which adds up over 400 files). I don't notice any issue myself but I have a nvme SSD which is pretty much the best case scenario for that.

I understand the desire to make everything "as optimal as possible", but this is the type of optimization that should probably only be done by a compiler/transpiler if it's wanted. It seems fairly simple to statically analyze a program and localize all repeated global variable accesses, and you might actually get a measurable improvement when doing this kind of optimization across all of the code and in more impactful places. I did this optimization in the gamemode's animations.lua, which is full of hooks that run for every player every tick, and improved the server tickrate by something like 10%. But the context of this function is not impactful at all.

So this change obviously doesn't have any significant upside in performance, but why do I dislike it? It's bloat - it makes the code more confusing. When you see _G getting set, you know you're setting globals, but when you see "environment" getting set, it takes extra analysis to figure out that "environment" is actually just globals. It makes the underlying program/AST more complicated and makes the file longer and take longer to read. Obviously for just this singular change, the negative impact is minor, but what it demonstrates is that you're not trying to make the code as simple as possible. Efficiency doesn't just mean absolute maximum performance when the margin of improvement is so insignificant; optimizing your program to be short and easier to understand is a more important objective.

swampservers:
Thanks, email me how you want payment. I can give you something else to do if you want if you read my comment https://github.com/swampservers/contrib/pull/255#discussion_r658425567

lua/autorun/client/cl_devtools.lua
NoSharp: I understand your response and will remember that for the future.

NoSharp:
@swampservers I sent you an email with the payment info.

PR #259, opened 2021-06-20
Unstick chatcommand! Will attempt to place you in the nearest standable spot. just type !stuck.

PYROTEKNIK:


lua/autorun/player_extension.lua
```diff
@@ -155,4 +155,111 @@ function PLAYER:UsingWeapon(cls)
...
+    end
+    tr.mins = mins
+    tr.maxs = maxs
+    local trace = util.TraceHull(tr)
+    if(trace.StartSolid)then return true end
+    return false
+end
+
+if (SERVER) then
+    concommand.Add("stuck", function(ply, cmd, args)
```
swampservers: What if this gets spammed?

PYROTEKNIK: The command is throttled, but I could probably optimize this further

PYROTEKNIK: wait no that's not supposed to still be there

swampservers: i mostly mean make sure there is a rate limit on it (10 sec cooldown)

lua/autorun/player_extension.lua
```diff
@@ -155,4 +155,103 @@ function PLAYER:UsingWeapon(cls)
     local c = self:GetActiveWeapon()
 
     return IsValid(c) and c:GetClass() == cls
+end
```
swampservers: Can you just put all this in chatcommands/stuck.lua please?

PYROTEKNIK: yeah one sec

lua/autorun/player_extension.lua
PYROTEKNIK: well i use whatever throttling fedorachat has. it also refuses to do any of the unstick functions if it doesn't think you're stuck

PR #261, opened 2021-06-20
Ammo HUD + SWEP improvements.

PYROTEKNIK:
Each change has been documented separately in each commit. 
The ammo hud should function pretty similar to gmod's stock one, it's just smaller and renders the icon behind the display. I would suggest merging this at the same time as the store weapon fix


lua/autorun/client/cl_ammo.lua
```diff
@@ -0,0 +1,131 @@
...
+local LASTAMMOTEXT = {}
+local LASTAMMOICON
+local grey = 25
+local grey_ab = 200
+
+hook.Add("HUDPaint","SwampHealthAmmo",function()
+
+    if(GetConVar("cinema_hideinterface") and GetConVar("cinema_hideinterface"):GetBool() == true)then return end
+    local ply=  LocalPlayer()
+    local drawhealth = true or ply:Alive() and ply:Health() < ply:GetMaxHealth()
```
swampservers: "true or"...?

PYROTEKNIK: sorry, left that in there by mistake

swampservers:
I'm not a fan of this design.

First of all, we don't need autoicons cluttering the screen. That's what the viewmodel is for.

Another thing we lose compared to the vanilla HUD is the scaling with screen size. I use Trebuchet24 for minor stuff out of laziness, but important things like the chatbox have screen scaling implemented and so should this.

Speaking more generally, i'm against custom HUDs and prefer HL2 style stuff because I think it makes the game more familar to people. How do you feel about making an HL2-style health/ammo display, and make it so both health and ammo are in the lower right so we can dedicate the left side to the chatbox?

HL2 built-in fonts: https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/sp/game/mod_hl2/resource/ClientScheme.res
(Use HalfLife2 font for numbers. Also note the blur and scanlines used for the number background)

Custom display should probably use: https://wiki.facepunch.com/gmod/WEAPON:CustomAmmoDisplay

PYROTEKNIK:
I can probably do that

PYROTEKNIK:
I kinda rushed my previous response. I designed this hud with the thought in mind that there was a reason you never just re-enabled the regular HL2 hud, which cinema disables by default. We don't really "lose" anything because the vanilla hud has never been there.
I personally think half life 2's is hud too big for a game that makes heavy use of other onscreen elements. 
As for the argument for familiarity, I would argue that most people who play garry's mod rarely play on sandbox or haven't played very much half life 2. Nearly all Gmod servers use some kind of custom interface. With that in mind i feel like you have the freedom to do something slightly better, and since this is swamp cinema, something simple and unobtrusive seems more logical than the half life 2 one. The inclusion of the autoicon is meant to make the player aware right away what the numbers they're looking at are associated with. I made my best attempt to make it not interfere with the player's view

I'm open to critique about the design for the hud i submitted. I can try something nicer for the design, I can probably even implement scaling. I just really don't get the point of trying to make it look like the hl2 hud. I think having the hud behave the same as hl2's, but take up less space is probably the better move design-wise.

swampservers:
I do use the vanilla HUD for ammo, weapon switcher, and crosshairs. The ammo just doesn't show for most weapon types.  I don't really want to deviate from the HL2 style as I do think it makes the game more familiar.

What I think we should do is have a Lua HUD that's designed to look very similar to the HL2 one, but with the health on the right side along with the ammo (since the chatbox is big on the left side), **and a bit smaller**. It still needs to scale with screenres, probably use https://wiki.facepunch.com/gmod/Global.ScreenScale . The HL2 hud says "HEALTH" and "AMMO" in a smaller font, I think that's much better than drawing the icon. The autoicon won't look good at a size other than 256x256 due to pixel sampling.

PYROTEKNIK:
That seems logical i just don't want to have to force it to be a big ugly yellow ammo counter just because everything else is the same big ugly yellow. Can i make it utilize the functions that read from clientscheme so maybe it's at least customizable on the user end?

PYROTEKNIK:
https://wiki.facepunch.com/gmod/Global.NamedColor

swampservers:
Merged in https://github.com/swampservers/contrib/pull/267, thank you!

PR #262, opened 2021-06-20
Dodgeball fix

PYROTEKNIK:
I know this isn't the timers thing you want but it looks really damn smooth even with latency i promise


lua/weapons/weapon_dodgeball/shared.lua
```diff
@@ -4,63 +4,139 @@ SWEP.PrintName = "Dodgeball"
...
+SWEP.Category = "Swamp Cinema" --todo remove these later i just need  this for testing 
 local outie = 64
 local innie = 28
 
+function SWEP:SetupDataTables()
+    self:NetworkVar("Int", 0, "ThrowState")
+    self:NetworkVar("Int", 1, "ThrowPower")
+    self:NetworkVar("Bool", 0, "Throwing")
+    self:NetworkVar("Float", 0, "StateTime")
+    self:NetworkVar("Entity", 0, "ThrownBall")
```
swampservers, line 20: This structure surely has duplicate data. Why do you need "Throwing" and "Throwstate" at the same time, can't you just use one value of Throwstate to indicate not thowing? you can even define GetThrowing to internally use the ThrowState. Alternately you could use StateTime=0 or ThrowPower=0 to indicate not throwing

The reason to change this isn't a marginal performance improvement, it's because it is more of a single-source-of-truth which is a core principle in software engineering to reduce bugs. With the current setup there might be a bug where Throwing gets set but not ThrowState and causes some kind of weird issue.

lua/weapons/weapon_dodgeball/shared.lua
```diff
@@ -4,63 +4,139 @@ SWEP.PrintName = "Dodgeball"
...
+
+        return
+    end
+
+    --play throw animation
+    if (curstate == 1) then
+        --self:EmitSound("Weapon_Pistol.Empty")
+        self:ThrowBall(self:GetThrowPower())
+        self:SetStateTime(CurTime() + 0.3)
+        self.Thrown = true
```
swampservers: unused?

lua/weapons/weapon_dodgeball/cl_init.lua
```diff
@@ -5,7 +5,21 @@ SWEP.Instructions = "Primary: Throw hard\nSecondary: Throw soft\nReload: Pass"
 SWEP.DrawAmmo = false
 SWEP.ViewModelFOV = 85
 
+function SWEP:ThrownBallExists()
+    local ply = self:GetOwner()
+    if(IsValid(self:GetThrownBall()))then return true end
+    if(IsValid(self.fakethrownball))then return true end
+    
+    local lookent = ply:GetEyeTrace().Entity
```
swampservers, line 13: So if they look at a dodgeball across the room the one they are holding dissapears? Why not just use a networked entity? Does the NW/DT ent not update in the same frame or something?

PYROTEKNIK: This is meant to be a quick way of getting the ball you threw as soon as its transmitted to you, since thrownball can be networked late, and that creates a small period of time where two dodgeballs are onscreen at once. I guess I could have it ignore balls with a  networked owner other than the player.

lua/weapons/weapon_dodgeball/shared.lua
PYROTEKNIK: Right. I will look at this very soon

lua/weapons/weapon_dodgeball/shared.lua
PYROTEKNIK: The sounds?

swampservers: Self.Thrown

lua/weapons/weapon_dodgeball/cl_init.lua
swampservers: Probably just do a range check

swampservers: The owner doesnt get unset when its sitting on the ground

lua/weapons/weapon_dodgeball/cl_init.lua
```diff
@@ -5,7 +5,21 @@ SWEP.Instructions = "Primary: Throw hard\nSecondary: Throw soft\nReload: Pass"
...
+    if(IsValid(lookent) and lookent:GetClass() == "dodgeball" and lookent:GetPos():Distance(ply:EyePos()) < 128)then
+        self.fakethrownball = lookent
```
swampservers, line 15: so if you're carrying a dodgeball, then walk up to another dodgeball on the ground and look at it, then the one you're holding turns invisible forever?

PYROTEKNIK: It should only use this value once thrown. Is there not a check?

PYROTEKNIK: looks like there's not

PR #284, opened 2021-07-18
Crusader sword in shop

PYROTEKNIK:
I feel like this weapon is too much fun to just leave it gone forever. I think this would be a good prize for the crusaders since they won and their original reward was rescinded.


swampservers:
Need to remove SetupMove (or refactor it with the flappy fedora - make SWEP:SetupMove())

PYROTEKNIK:
Good idea. I will work on that

PYROTEKNIK:
sorry about the pissing around, i didn't realize you had implemented it already

swampservers:
You have a KeyRelease hook in here that does literally nothing and the check is run every time any player presses any key.

PYROTEKNIK:
> You have a KeyRelease hook in here that does literally nothing and the check is run every time any player presses any key.

there was a float jump thing in there. did you remove it during the war or was it me?

swampservers:
> > You have a KeyRelease hook in here that does literally nothing and the check is run every time any player presses any key.
> 
> there was a float jump thing in there. did you remove it during the war or was it me?

Maybe IDK, either way, it looks like you deleted all of the jump stuff which was a good idea

PR #286, opened 2021-07-18
Slitter Collision BBox [Choose this or Basic Slitter Fix]

PYROTEKNIK:
The last update apparently introduced some issues with the slitter, which caused it to only hit players if you aim inside their capsule but not at their actual body

I did this whole change without realizing it probably happened because i changed it to use Ents["Player"] as the trace filter instead of player.GetAll(). Maybe util.TraceLine doesn't like non-sequential filter tables?

Anyway i think this is probably better anyway. I Changed the collision testing to use collisionbounds bbox instead of that weird capsule  check. This does take into account crouching, but i can change that and use a static bbox instead if you want. I think testing bbox is probably a lot cleaner, but this is your call, and i don't even know if the other one will solve the problem.

### DIFF
--- a/lua/weapons/weapon_slitter/shared.lua
+++ b/lua/weapons/weapon_slitter/shared.lua
@@ -133,32 +133,61 @@ function SWEP:TraceCapsule(vp, vr, cp, ca, cmin, cmax, radius, maxdist)
     return self:TraceSphere(vp, vr, cp + thing * ca, radius, maxdist)
 end
 
+--can we just do this instead of some gay capsule please it's way simpler
+function SWEP:TestHit(target)
+    local ply = self:GetOwner()
+    local bbox_origin = target:GetPos()
+    local mins, maxs = target:GetCollisionBounds() --or we can use a static bbox
+    local bscale = Vector(0.7, 0.7, 1)
+    mins = mins * bscale
+    maxs = maxs * bscale
+    local ray_origin = ply:GetShootPos()
+    local ray_delta = ply:GetAimVector() * 100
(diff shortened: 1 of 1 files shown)


swampservers:
Capsule fits the player shape better and there's nothing wrong with it. It could easily be changed to support crouching. Using this :GetCollisionBounds() is a potential unfairness issue with different models.

PYROTEKNIK:
Oh the point wasnt to support crouching the point was that Cubes are just a lot cheaper for the engine to test

swampservers:
Oh sure dude, it's gonna be way cheaper when your TestHit function gets checked for each player and then within TestHit, it calls player.GetAll() each time (even if it's not going to use it, because you pointlessly use the allply variable outside the if statement) making your check O(N^2) overall.

Show some actual evidence that util.IntersectRayWithOBB is faster because that sounds like pure conjecture and I bet my ray/capsule intersection code is way faster anyway.

You obviously made this because your change broke the slitter and rather that doing the obvious thing and looking at which lines were changed when the break occurred, you perceived a possible problem in code I wrote years ago and has never had an issue before and decided to rewrite it because you didn't understand the math in it.

PYROTEKNIK:
I think you're very severely misunderstanding why I made this one. I never blamed the way you set it up for the issue that came up. I was just at a loss for why this happened in the first place. I'm sure capsules are fine, I just figured it would be faster to do bounding box collision because it's functionally identical to the way the engine handles all other hitscan between characters. I quickly whipped up the other branch afterward because I was running out of time to work on this, and hadn't done what I wanted to with it, and wanted to gave a discussion with you about it before doing anything else, since this was mainly just an optimization idea I had while I was looking at the problem.

swampservers:
Benchmark it, then

PR #287, opened 2021-07-18
Basic Slitter Fix

PYROTEKNIK:
Apparently Ents["player"] doesn't work right in trace filters.
I don't know if this will actually solve the problem, this is just my best guess.


swampservers:
Yeah, Ents.player isn't a sequential table.

PYROTEKNIK:
Well yeah I know that I just didn't think it would matter for trace.filter

PR #289, opened 2021-07-25
swep error checks

PYROTEKNIK:


lua/weapons/weapon_flare.lua
```diff
@@ -51,34 +51,36 @@ function SWEP:PrimaryAttack()
...
-        local effectdata = EffectData()
-        effectdata:SetOrigin(org)
-        effectdata:SetNormal(dir)
-        effectdata:SetRadius(1)
-        effectdata:SetMagnitude(1)
-        effectdata:SetScale(0.1)
-        util.Effect("ElectricSpark", effectdata)
-
-        --[[ local hit = self.Owner:GetEyeTrace()
+        if (IsValid(ply)) then
```
swampservers, line 54: Have you seen this error happen? That seems like a bug in the entitytimer system if so

PYROTEKNIK: there was an error about a null entity in the flare swep on the server but it said it was on line 3. i wasn't sure where else it could have occurred

swampservers: Through serverdebug?

PYROTEKNIK: yeah

swampservers: I think the problem is actually the `self` becoming invalid, because the ply:TimerSimple won't callback if ply is invalid.

swampservers:
merging becuase i just got the fuckin monke weapon error again and it needs to end

PR #290, opened 2021-07-26
Native Linux Input Workaround

Noztik:
https://pastebin.com/BhvF3MCG
please replace the conditional in SwampChat:Open with this, fixing the chatbox input is more important than fixing the request browser input


swampservers:
https://wiki.facepunch.com/gmod/system.IsLinux

Noztik:
> https://wiki.facepunch.com/gmod/system.IsLinux

it should be a concommand so if they fix the issue I don't have to bug you to remove the fix

PR #292, opened 2021-07-28
Add player duel command

Evaisie:
A duel command that lets players 1v1 eachother for points.
When a player wins they take the points from the other player.

The commands are
`!duel [player] [amount of points] [number of kills to win]`
`!duel accept [confirm number of points]`

Note that since I do not have a testing environment which matches Swamp cinema I did not test how any of the messages look in the real swamp chat.
I basically learned the Swamp cinema specific functions from the coinflip command.


lua/chatcommands/duel.lua
```diff
@@ -0,0 +1,284 @@
...
+        if (v.player1 == playerID or v.player2 == playerID) then
+            return true
+        end
+    end
+
+    return false
+end
+
+function requestDuel(player1, player2, points, kills)
+    if (not hasPendingDuel(player1:UserID(), player2:UserID(), points)) then
```
swampservers: Why bother with these UserIDs? Just use the actual player object

Evaisie: Oh whoops, that is a leftover from me testing, something was acting up when I was testing with multiple clients running under the same steam account (Or perhaps testing with bots). I'll clean it up later.

I noticed that I forgot to uncomment something else too

Evaisie: There that should be a bit better

lua/chatcommands/duel.lua
```diff
@@ -0,0 +1,271 @@
...
+            else
+                player1:ChatPrint("[red]The win condition can not be more than 20 kills.")
+            end
+        else
+            player1:ChatPrint("[red]Player " .. string.Implode(" ", args) .. " not found.")
+        end
+    end
+end)
+
+function hasPendingDuel(player1, player2, points)
```
swampservers: This code would be a lot better if the 2 players just got sorted (by ent index maybe)

lua/chatcommands/duel.lua
```diff
@@ -0,0 +1,271 @@
...
+            table.insert(NewDuelRequests, v)
+        end
+    end
+
+    DuelRequests = NewDuelRequests
+    local ActiveDuelsNew = {}
+    for k, v in pairs(ActiveDuels) do
+
+        if (v.player1_points >= v.kills) then
+            v.player2:SS_TakePoints(amount)
```
swampservers: What if the losing player spends or transfers their points before the duel is over?

EvaisaGiac: Hmm, taking the points at the start of the duel from both sides would work, although it would be shitty if the server crashes or something while the duel is going on.

Evaisie: Hmm, taking the points at the start of the duel from both sides would work, although it would be shitty if the server crashes or something while the duel is going on.

lua/chatcommands/duel.lua
Evaisie: Hmm I _think_ I understand what you mean, lemme see

Evaisie: hmm no I don't actually know what you mean by that, I thought you meant like a filter or something but I don't think that would make the code smaller, I do agree its kind of ugly

Evaisie: I'm gonna go through the entire code and clean it up, I found a few mistakes already, one thing that I can do better is variable names, also probably use more modular functions.

Evaisie (commented):
I split a few of the messy functions into more modular less cluttered functions to that the script is more readable.
I added a duels command which shows all the duel requests you currently have.
I changed a bunch of variable names to some which make more sense.
`player1 -> challenger` and `player2 -> target` for example.
I also moved the part where it takes the points from the players to when the player accepts the duel. This way they can not give points away or lose points while the duel is going on.
One issue with this however is that if the server crashes while the duel is going on the points are most likely lost to the void. Not sure how big a deal this is.

lua/chatcommands/duel.lua
```diff
@@ -0,0 +1,313 @@
...
+    local duel = AcceptedDuels[duelIndex]
+    local time = 20
+    timer.Create( "CountdownDuel" .. duelIndex, 1, 20, function()
+        time = time - 1
+
+        if (not IsValid(duel.challenger)) then
+            duel.target:SS_GivePoints(duel.points * 2)
+            duel.target:ChatPrint("[red]Your opponent " .. duel.challenger_name .. " has left the game. [green]You won [gold]" .. string.Comma(duel.points) .. "[green] points.")
+            BotSayGlobal("[fbc]" .. duel.target_name .. " won a duel worth [rainbow]" .. string.Comma(duel.points * 2) .. " points [fbc] against [gold]" .. duel.challenger_name .. "[fbc]. Their opponent left the game.")
+            AcceptedDuels[duelIndex] = nil
```
swampservers: This isn't how you remove something from a sequential table. There will be a "gap" in the table. table.remove will remove it and shift other values down, but that will cause problems with other duels when their indexes change.

At first glance I can already confirm that this would cause massive issues when combined with the startCountdown call on line 234. The # operator on tables returns *any index* such that `t[i] ~= nil and t[i+1] == nil`. This will behave strangely if there are multiple such indices in the structure. You might table.insert a new duel (line 231) into index 10, which fills a gap in the middle of the table, and then the # call on line 234 would pick index 20, the actual end of the table (or another gap...) and you'd end up calling startCountdown on one duel multiple times while not calling it at all on another duel. This is probably totally exploitable by creating some duels with low point values and others with huge values and manipulating the system into taking the low point values away and paying out the high values (perhaps to different accounts, but this is almost definitely possible)

I'm gonna be honest with you, this code seems overcomplicated and I really don't trust it not to have other edge cases in it which could blow up my server economy. Why are you using sequential tables instead of key/values? Why are AcceptedDuels and ActiveDuels two separate tables?

lua/chatcommands/duel.lua
```diff
@@ -0,0 +1,313 @@
...
+        elseif (not IsValid(duel.target)) then
+            duel.challenger:SS_GivePoints(duel.points * 2)
+            duel.challenger:ChatPrint("[red]Your opponent " .. duel.target_name .. " has left the game. [green]You won [gold]" .. string.Comma(duel.points) .. "[green] points.")
+            BotSayGlobal("[fbc]" .. duel.challenger_name .. " won a duel worth [rainbow]" .. string.Comma(duel.points * 2) .. " points [fbc] against [gold]" .. duel.target_name .. "[fbc]. Their opponent left the game.")
+            AcceptedDuels[duelIndex] = nil
+            timer.Remove( "CountdownDuel" .. duelIndex )
+            return
+        end
+
+        if (time == 15) then
```
swampservers: make this shit less redundant 

```
if ({15=true,10=true,5=true,4=true,3=true,2=true,1=true})[time] then
    for i,ply in ipairs({duel.challenger,duel.target}) do
        ply:ChatPrint("[orange]Duel starts in " .. time .. " seconds")
    end
end
```

swampservers: all the other places you do the same thing for both players should also loop over them in the same way

lua/chatcommands/duel.lua
```diff
@@ -0,0 +1,313 @@
...
+        duel.target:ChatPrint("[red]You do not have enough points to accept this duel.")
+        duel.challenger:ChatPrint("[red]The challenged player does not have enough points, duel was cancelled.")
+        DuelRequests[duelIndex] = nil
+        return
+    end
+
+    duel.challenger:ChatPrint("[orange]" .. duel.target_name .. " accepted your duel request, duel will start in 20 seconds!")
+    duel.target:ChatPrint("[orange]Your duel with " .. duel.challenger_name .. " will start in 20 seconds!")
+
+    duel.challenger:SS_TakePoints(duel.points)
```
swampservers: So in order to prevent exploits with multi-logging on different servers (which will be a thing in the near future), you actually need to use SS_TryTakePoints. This will atomically check if the player actually has enough points, and then take them if so, and call one callback or the other depending on if it worked.
```
ply:SS_TryTakePoints(amount, function() print("It worked") end, function() print("It failed") end)
```
Doing this with two players is a bit of a hassle, since if you charge guy A but guy B fails to charge you need to refund guy A before canceling the duel.

lua/chatcommands/duel.lua
Evaisie: Alright I will take all your critique in mind and do a full revision. I am getting used to Lua again and I think I can do better. Thanks for all the input, it helps a lot.

lua/chatcommands/duel.lua
Evaisie: Well I learned something today, I have never seen someone use a table this way

lua/chatcommands/duel.lua
Evaisie: Is this any different from just checking if both the player have enough points and then taking the points?

swampservers: The problem is that the actual number of points lives on a database which is accessed asynchronously potentially by multiple servers/processes. The TryTakePoints function will atomically try to take the points away and call the callback only on success.

EvaisaGiac:
Any update on this?

Evaisie:
Any update on this?

Kefta:
Reviewing this today

PR #293, opened 2021-07-29
Fix trash Lua error

Evaisie:
Trash was causing lua errors when holding the tape because there was a null check missing for the Trashfriends table.


swampservers:
Thanks, I already fixed it though

PR #299, opened 2021-09-08
Contribution

RossCattero:
For trivia system;


lua/autorun/sh_netstream.lua
```diff
@@ -0,0 +1,164 @@
...
+
+					bShouldSend = true;
+				end;
+			end;
+		else
+			bShouldSend = true;
+		end;
+
+		local dataTable = {...};
+		local encodedData = pon.encode(dataTable);
```
swampservers, line 82: just net.WriteTable...?

lua/autorun/sh_netstream.lua
```diff
@@ -0,0 +1,164 @@
+--[[
```
swampservers: you don't need this shit or sh_pon, use built in networking

lua/autorun/client/cl_ross_derma.lua
```diff
@@ -0,0 +1,152 @@
...
+
+        if self.bColor then
+            surface.SetDrawColor(Color(40, 40, 40))
+            surface.DrawOutlinedRect( 0, 0, w, h, 1 )
+        end;
+    end;
+end;
+
+// I've taken it from the helix;
+function BetterWrapText(text, maxWidth, font)
```
swampservers: Use my draw.WrappedText https://github.com/swampservers/contrib/blob/20f52368d7e9135be53213934ff3b5b6f96f811e/lua/autorun/client/stylesheet.lua

Do not make your code larger than it needs to be!!!! excess/duplicate code is a maintenance nightmare

lua/autorun/client/cl_votestuff_init.lua
```diff
@@ -0,0 +1,77 @@
...
+surface.CreateFont( "Question-font", {
+	font = "Century Gothic",
+	size = ScreenScale(7),
+	scanlines = 0,
+	antialias = true,
+	extended = true,
+})
+
+surface.CreateFont( "Anwser-font", {
+	font = "Century Gothic",
```
swampservers, line 14: is this font on people's PC?

lua/autorun/sh_votestuff.lua
```diff
@@ -0,0 +1,25 @@
+// Shared utils for votestuff by Ross;
+
+function RFormatTime(time)
```
swampservers: https://wiki.facepunch.com/gmod/string.FormattedTime

lua/autorun/client/cl_ross_derma.lua
```diff
@@ -0,0 +1,152 @@
...
+    end
+
+    if (line != "") then
+        lines[#lines + 1] = line
+    end
+
+    return lines
+end
+    
+function panelMeta:Close()
```
swampservers: i would rather not put this in, it might cause issues and its totally out of scope of this change. 

(meaning, this whole file, should be removed, and i am just commenting here as an example of something that seems like a potential glitch)

swampservers (changes requested):

lua/autorun/sh_netstream.lua
RossCattero: Okie. sh_pon and netstream just makes code looks more compact, but as you wish.

lua/autorun/client/cl_ross_derma.lua
RossCattero: Ok, I'll use yours instead.

Where is it larger than it should be?

lua/autorun/client/cl_votestuff_init.lua
RossCattero: Yes, it is. It's the default font on every **windows** PCs. For linux or MAC it will use the default fonts for their system.

lua/autorun/sh_votestuff.lua
RossCattero: > https://wiki.facepunch.com/gmod/string.FormattedTime

I've done this function before this appeared in gmod I think. I'll use it, thanks for point

lua/autorun/client/cl_ross_derma.lua
RossCattero: > i would rather not put this in, it might cause issues and its totally out of scope of this change.
> 
> (meaning, this whole file, should be removed, and i am just commenting here as an example of something that seems like a potential glitch)

Agreed. I left this file for future purposes if I'll make a new VGUIs in future. But I'm going to remove it

lua/autorun/server/sv_votestuff.lua
```diff
@@ -0,0 +1,53 @@
...
+	10,
+} 
+GivePoints = 100; // Amount of points to give the winner;
+Curr_Task = Curr_Task || {}
+
+// Creating a random timer on init;
+local uniqueID = 'FetchQuestions'
+local time = Seconds[math.random(1, #Seconds)]
+timer.Create(uniqueID, time, 0, function()
+	if !timer.Exists(uniqueID) then timer.Remove(uniqueID) return; end;
```
swampservers, line 16: line is unused (and makes no sense, you're removing something you know doesn't exist)

lua/autorun/client/cl_votestuff_init.lua
```diff
@@ -0,0 +1,73 @@
+if SERVER then return end;
```
swampservers, line 1: won't happen, why leave this check in?

lua/autorun/client/cl_votestuff_init.lua
```diff
@@ -0,0 +1,73 @@
...
+
+local filterPattern = "&#?[%w]+;"
+
+surface.CreateFont( "Question-font", {
+	font = "Century Gothic",
+	size = ScreenScale(7),
+	scanlines = 0,
```
swampservers, line 8: generally you shouldn't specify parameters that match the default values, and certainly not for this

lua/autorun/client/cl_votestuff_init.lua
```diff
@@ -0,0 +1,73 @@
...
+surface.CreateFont( "Anwser-font", {
+	font = "Century Gothic",
+	size = ScreenScale(6.5),
+	scanlines = 0,
+	antialias = true,
+	extended = true,
+})
+
+net.Receive("tv::SendTasks", function( len )
+		local time = net.ReadUInt(16);
```
swampservers, line 22: You don't need to change it as I have a thing that will fix this automatically, but for future reference - you're not using standard lua syntax, you're using retarded Garry syntax, which is considered amateur

- don't put ; at the end of lines
- use "not" instead of !
- use "and" instead of &&
- use "or" instead of ||

lua/autorun/client/cl_votestuff_init.lua
```diff
@@ -0,0 +1,73 @@
...
+					net.Start("tv::SendAnswer")
+						net.WriteString( tasks.answers[num] )
+					net.SendToServer()
+
+					surface.PlaySound("buttons/button15.wav")
+				end;
+				return true;
+		end, 
+		function() 
+			local lFont = "Question-font"
```
swampservers, line 40: these variables are frivolous just put the string into the function call

lua/autorun/client/cl_votestuff_init.lua
```diff
@@ -0,0 +1,73 @@
...
+
+surface.CreateFont( "Question-font", {
+	font = "Century Gothic",
+	size = ScreenScale(7),
+	scanlines = 0,
+	antialias = true,
+	extended = true,
+})
+
+surface.CreateFont( "Anwser-font", {
```
swampservers, line 13: misspelled

lua/autorun/client/cl_votestuff_init.lua
```diff
@@ -0,0 +1,73 @@
+if SERVER then return end;
+
+local filterPattern = "&#?[%w]+;"
```
swampservers, line 3: this should be on line 45

swampservers: i have literally no clue what this is for outside of its proper context

lua/autorun/server/sv_votestuff.lua
```diff
@@ -0,0 +1,53 @@
...
+		timer.Adjust(uniqueID, time + Seconds[math.random(1, #Seconds)], 0, nil)
+end);
+
+net.Receive("tv::SendAnswer", function( len, ply )
+		if !ply:IsValid() || !Curr_Task.correct then return end;
+		local anwser = net.ReadString()
+		local correct = Curr_Task.correct == anwser;
+
+		if correct then
+				-- client:SS_GivePoints( GivePoints )
```
swampservers, line 47: if this code was enabled, a hacker could just spam this net message and give himself infinite points... come on man

lua/autorun/server/sv_votestuff.lua
```diff
@@ -0,0 +1,53 @@
...
+				net.WriteTable(Curr_Task) // Sending a table; In my opinion faster is sending a JSON string instead of table itself;
+			net.Broadcast() // Sending to everyone;
+	end)
+	
+		math.randomseed( os.time() );
+		timer.Adjust(uniqueID, time + Seconds[math.random(1, #Seconds)], 0, nil)
+end);
+
+net.Receive("tv::SendAnswer", function( len, ply )
+		if !ply:IsValid() || !Curr_Task.correct then return end;
```
swampservers, line 42: ply will always be valid

lua/autorun/server/sv_votestuff.lua
```diff
@@ -0,0 +1,53 @@
+if !SERVER then return end
+
+util.AddNetworkString( "tv::SendTasks" )
+util.AddNetworkString( "tv::SendAnswer" )
+
+Seconds = { // Each X seconds will appear a vote for all online players. They will have X/2 seconds to anwser;
+	10,
```
swampservers, line 7: So you have logic lower down for selecting a random number from this table, but there is only one value in the table. This code is obviously unfinished and i'm not interested in finishing it for you.

Also "Seconds" is an inappropriate name for a global variable; it is likely to conflict with something else with equivalently careless variable naming.

lua/autorun/server/sv_votestuff.lua
```diff
@@ -0,0 +1,53 @@
...
+} 
+GivePoints = 100; // Amount of points to give the winner;
+Curr_Task = Curr_Task || {}
+
+// Creating a random timer on init;
+local uniqueID = 'FetchQuestions'
```
swampservers, line 13: frivolous variable

lua/autorun/server/sv_votestuff.lua
```diff
@@ -0,0 +1,53 @@
+if !SERVER then return end
```
swampservers, line 1: frivolous check

swampservers:
Sorry about the delay in reviewing this

PR #310, opened 2021-10-26
Fix pony weapon positions for spray paint and stencil paint

ghost:


swampservers:
Thanks, but this isn't the correct way to fix this. The file you're changing is meant to override model positions for HL2 weapons only. You want to edit weapon_spraypaint.lua because it's a SWEP.

You can use the devtools (linked in readme) to test on the server.

ghost:
Thanks, will do. I think I can just test this locally for now

ghost:
Okay I looked at other sweps like popcorn for how to do this. For my local instance, the position wasn't correct for human either, but I think it is correct on the server. I set it here, let me know if I should only set it for pony. I'm not sure how it works on the server

swampservers:
Thanks, we'll see what it does. You can go back and edit it using the devtools on the live server.

PR #319, opened 2021-11-30
Age Restriction Bypass

Noztik:


gamemodes/cinema/gamemode/modules/theater/services/sh_youtube.lua
```diff
@@ -30,22 +30,125 @@ function SERVICE:GetKey(url)
...
+					YOUTUBE_TRUE_START_PING = SysTime()
+				end
+			end
         end
     end
+	
+    function SERVICE:SetVolume(vol, panel)
+		if (self:IsMature(panel.vid)) then
+			local nvar = RandomString()
+			local str = "var "..nvar.."=setInterval(function(){if(typeof(th_volume)!='undefined'){th_volume("..vol..");clearInterval("..nvar..");}},10);"
```
swampservers: You're using a random variable name so that if this gets runs multiple times, each run will store its value in a different variable. Try using a scope instead https://www.w3schools.com/js/js_scope.asp

gamemodes/cinema/gamemode/modules/theater/services/sh_youtube.lua
```diff
@@ -30,22 +30,125 @@ function SERVICE:GetKey(url)
...
+			enabled = true
+			loaded = false
+			timer.Create(RandomString(), .1, 30, function()
+				if IsValid(panel) then
+					if enabled then
+						panel:RunJavascript(payload)
+						panel:RunJavascript("if(window.ytcfg)console.log('PAGE LOADED');")
+						self:SetVolume(theater.GetVolume(), panel)
+					end
+					panel:RunJavascript([[
```
swampservers, line 54: This doesn't belong here, it belongs in the html file.

Perhaps we should stop using remote html files for theater services and just panel:SetHTML?

gamemodes/cinema/gamemode/modules/theater/services/sh_youtube.lua
```diff
@@ -30,22 +30,125 @@ function SERVICE:GetKey(url)
...
...entRenderer.items.find((e=>e.expandableVideoDescriptionBodyRenderer));c.expandableVideoDescriptionBodyRenderer&&(l.expandableVideoDescriptionBodyRenderer=c.expandableVideoDescriptionBodyRenderer)}try{attachInitialDataInterceptor(checkAndUnlock),attachJsonInterceptor(checkAndUnlock),attachXhrOpenInterceptor(onXhrOpenCalled)}catch(e){error(e,"Error while attaching data interceptors")}function checkAndUnlock(e){try{isPlayerObject(e)&&isAgeRestricted(e.playabilityStatus)?unlockPlayerResponse(e):isPlayerObject(e.playerResponse)&&isAgeRestricted(e.playerResponse.playabilityStatus)?unlockPlayerResponse(e.playerResponse):isEmbeddedPlayerObject(e)&&isAgeRestricted(e.previewPlayabilityStatus)?unlockPlayerResponse(e):isWatchNextObject(e)&&isWatchNextSidebarEmpty(e)?unlockNextResponse(e):isWatchNextObject(e.response)&&isWatchNextSidebarEmpty(e.response)&&unlockNextResponse(e.response)}catch(e){error(e,"Video or sidebar unlock failed")}return e}function onXhrOpenCalled(e,t,n){if(isGoogleVideo(t,n))return isGoogleVideoUnlockRequired(n,getLastProxiedGoogleVideoId())?(Object.defineProperty(e,"withCredentials",{set:()=>{},get:()=>!1}),getGoogleVideoUrl(n.toString(),VIDEO_PROXY_SERVER_HOST)):void 0}
```
swampservers, line 34: wtf is this

gamemodes/cinema/gamemode/modules/theater/services/sh_youtube.lua
Noztik: there's not really any difference between using a webpage and using sethtml except for latency however in this case I _need_ to have it open the youtube embed page due to web protection crap like CORS
I would love to have it all in the remote webpage if I could but I can't due to iframe bs and function overriding race conditions that the bypass needs

gamemodes/cinema/gamemode/modules/theater/services/sh_youtube.lua
Noztik: a userscript I copied and minimized that allows this bypass, it's more work than it's worth to shift through it and trim it

gamemodes/cinema/gamemode/modules/theater/services/sh_youtube.lua
swampservers: Oh I see, I didn't notice that you were changing the page URL. This is good then.

PR #322, opened 2021-12-14
Fixes

Noztik:


gamemodes/cinema/gamemode/misc/cl_hl2hud.lua
```diff
@@ -49,6 +49,12 @@ function GM:PlayerStartVoice(ply)
     end
 end
 
+function GM:DrawDeathNotice(x, y)
```
swampservers: Just use a hook

lua/swamp/dev/cl_devtools.lua
```diff
@@ -78,6 +78,7 @@ local function loadFile(filePath, environmentVars)
     end
 
     xpcall(function()
+		GM = GAMEMODE
```
swampservers: pls set GM=nil after

gamemodes/cinema/gamemode/misc/cl_hl2hud.lua
Noztik: the death notice is already drawn by the time it gets to the hook I think, even just having it only contain "return" doesn't stop the death notices

swampservers: Try returning true. Returning nothing is nil (same as not having a return statement at all). You have to return something non-nil to "block" a hook. It looks to me like the death notice drawer is a gamemode function, so it would run only if any hook.Add functions don't block it.

PR #330, opened 2021-12-28
Update weapons for protection notify.

LegacyValken:
Inserts missing weapon classes for notification during a private theater that you cannot harm/can harm inside of the theater space.


gamemodes/cinema/gamemode/theater/sh_protect.lua
```diff
@@ -77,7 +77,45 @@ if CLIENT then
...
         weapon_gauntlet = true,
-        weapon_peacekeeper = true
+        weapon_peacekeeper = true,
+        weapon_magicmissile = true,
+        weapon_garand = true,
+        weapon_pickaxe = true,
+        weapon_pickaxe_diamond = true,
+        weapon_fists = true,
+        weapon_crusadersword = true,
+        gun_44magnum = true,
```
swampservers: Just do class:StartWith("gun_")

LegacyValken: I didn't see this in the API or was even aware of string.StartWith exists. Will update accordingly.

swampservers:
I've applied these changes manually because I moved the files around so it was easier to just paste the code in. Thanks

PR #332, opened 2021-12-29
Small Title Changes

Metroid55:
Added a few titles
Did some cleanup
Commented out forced titles


swampservers:
I've made most of these changes manually, thanks

PR #333, opened 2022-01-03
fix count in swag tab in swampshop

bobolobxd:
The class for accessories is simply "accessory" rather than a unique name, so it counts all accessories in the same loop in the CountItems function. This change makes it so that if the class is 'accessory', it instead compares the model path so there is a unique count for all the other accessories.

before: https://i.imgur.com/zaMjiTI.png
after: https://i.imgur.com/vQjYq4S.png


swampservers:
Thank you! However, this would break sh_products.lua which also calls this CountItem function and hasn't been updated.

You can fix it if you want. FYI, though, the whole shop/item system is kinda the main thing i'm working on RN so i'll fix this issue sooner or later anyway

swampservers:
Nice trips btw

PR #334, opened 2022-01-05
LookMovie Fix

Noztik:
their captcha is gone although it was only up for a day, this should work unless they decide to put it back up


swampservers:
Thank you!

PR #335, opened 2022-01-07
Mute History

Noztik:


lua/cinema/scoreboard/cl_playerlist.lua
```diff
@@ -64,7 +86,7 @@ function UpdateMutes()
     end
 end
 
-timer.Create("updatemutes", 1, 0, UpdateMutes)
+timer.Create("updatemutes", 1, 0, UpdateMuteHistory)
```
swampservers: So you're gonna read and write back a file every second, which synchronously hangs the client? Please come up with a better solution

lua/cinema/scoreboard/cl_playerlist.lua
```diff
@@ -51,11 +51,33 @@ concommand.Add("mute", function(ply, cmd, args, argss)
...
-        UpdateMutes()
+        UpdateMuteHistory(v)
+        print(v.ClickMuted and " Muted" or " Unmuted", v)
     end
 end)
 
+function UpdateMuteHistory(ply)
+    local history = util.JSONToTable(file.Read("swamp_mutehistory.txt", "DATA") or "") or {}
+
+    if ply then
```
swampservers:
```
if ply then
    history[ply:SteamID()].mute = (ply.ClickMuted or ply.IsChatMuted) and {mute=ply.ClickMuted, chatmute=ply.IsChatMuted} or nil
end
```

lua/cinema/scoreboard/cl_playerlist.lua
Noztik: yea I didn't know what the impact was and I thought it was fine since the timer iterated through every player in the server before

lua/cinema/scoreboard/cl_playerlist.lua
```diff
@@ -51,11 +51,28 @@ concommand.Add("mute", function(ply, cmd, args, argss)
...
 end)
 
+local history = util.JSONToTable(file.Read("swamp_mutehistory.txt", "DATA") or "") or {}
+function UpdateMuteHistory(ply)
+    if ply then
+        history[ply:SteamID()] = (ply.ClickMuted or ply.IsChatMuted) and {mute=ply.ClickMuted, chatmute=ply.IsChatMuted} or nil
+    end
+
+    for k,v in pairs(history) do
+        local p = player.GetBySteamID(k)
```
swampservers: You have an `O(numplayers * historysize)` function here and it runs every time a new player gets created. 

You could easily change it to `O(numplayers)` by iterating all players and looking up their steamIDs in the table. The issue is player.GetBySteamID has to iterate all players.

But what you should actually do is factor your code properly so you're only changing the ClickMuted/IsChatMuted/SetMuted/whatever on individual players instead of resetting them all and rewriting your file every time a player enters PVS. Do not repeatedly set things if there is no reason to. So when the user clicks a mute button, all it does is update the mute state of THAT player and saves the file. Likewise when a player entity is created, only update THAT player and DONT save the file.

lua/cinema/scoreboard/cl_playerlist.lua
```diff
@@ -64,6 +71,16 @@ function UpdateMutes()
     end
 end
 
+hook.Add( "NetworkEntityCreated", "UpdateMuteHistory", function(ent)
+    if ent:IsPlayer() then
+		local p = history[ent:SteamID()]
+		if p then
+			if p.mute then ent.ClickMuted = true end
+			if p.chatmute then ent.IsChatMuted = true end
```
swampservers: Where is UpdateMutes called? Don't you need to ply:SetMuted also?

lua/cinema/scoreboard/cl_playerlist.lua
```diff
@@ -51,11 +51,18 @@ concommand.Add("mute", function(ply, cmd, args, argss)
...
-        UpdateMutes()
+        UpdateMuteHistory(v)
+        print(v.ClickMuted and " Muted" or " Unmuted", v)
     end
 end)
 
+local history = util.JSONToTable(file.Read("swamp_mutehistory.txt", "DATA") or "") or {}
+function UpdateMuteHistory(ply)
+    history[ply:SteamID()] = (ply.ClickMuted or ply.IsChatMuted) and {mute=ply.ClickMuted, chatmute=ply.IsChatMuted} or nil
+    UpdateMutes()
```
swampservers: so you still check every single player when only updating one

Noztik: that's literally how it is in the current code except it runs every second

PR #340, opened 2022-02-09
Prevent prop spawn while dead.

LegacyValken:
Got tired of others being able to spawn explosive props when dead. I'm not 100% sure if this will actually prevent it, but judging from how products work it seems like it might work.

Tell me otherwise if I can do this better.

### DIFF
--- a/lua/swamp/swampshop/sh_products.lua
+++ b/lua/swamp/swampshop/sh_products.lua
@@ -7,7 +7,7 @@ function SS_Product(product)
 
     function product:CannotBuy(ply)
         if ply.SQLCreatingItem then return "Database lock, try again." end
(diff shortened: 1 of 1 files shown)


lua/swamp/swampshop/sh_products.lua
```diff
@@ -7,6 +7,7 @@ function SS_Product(product)
 
     function product:CannotBuy(ply)
         if ply.SQLCreatingItem then return "Database lock, try again." end
+        if self.class == "prop_physics" then ply:Notify('Cannot spawn props while dead.') return end
```
swampservers: uh huh...

Okay so. You're trying to prevent something from being done while dead. In what place in your code do you check `player:Alive()` which would be logically mandatory to have a condition like that?

LegacyValken: Whoops, I forgot the alive function. Unless there is somewhere else you're suggesting I put this?

PR #343, opened 2022-02-15
another button fix

bobolobxd:
https://gfycat.com/everyalienatedhogget


lua/cinema/entities/magicbutton.lua
```diff
@@ -304,7 +304,11 @@ function ENT:Draw()
 
     render.SetAmbientLight(0.05, 0.05, 0.05)
     --render.SuppressEngineLighting( true )
-    self:DrawModel()
+    if util.IsValidModel("models/pyroteknik/secretbutton.mdl") then
+        self:DrawModel()
+    else
```
swampservers: the else/setnodraw is pointless. if the if check fails, you don't draw the model.

bobolobxd: I added it because it was still drawing a shadow when i was testing it on theatron

bobolobxd: https://i.imgur.com/AqyHB8w.png this is without setnodraw

PR #354, opened 2022-04-21
slight shotgun nerfs

Metroid55:
also made the spas12 shoot a bit slower
someone should look into balancing other weapons as well or maybe I will at some point idk


swampservers:
thank you

PR #360, opened 2022-05-28
Update sh_titles.lua

Blisskiwi:


swampservers:
There are a bunch of issues so I just did the change myself

PR #367, opened 2022-07-16
Fixes

Noztik:


swampservers:
Um, doesn't this break it? Can you run the code and then get a slitter that isn't using the default cleaver model and test that?

Noztik:
It seems to work fine when I just use the default slitter.

swampservers:
what about the other slitters?

You deleted the code that reads `SlitterModels` which contains their position/angle

Noztik:
I don't know, I only have the cleaver and it would be hard for me to get a different slitter. Shouldn't it be fine since the default slitter is part of the same system and it works fine?

swampservers:
removed these lines https://gyazo.com/eed1560f7174b8a0c02cc0f4e95d4eae

Should mention that the different models are all finicky in their own ways so I have to like set it twice

Noztik:
I put the slitter lines back, I still haven't tested it with other slitters because I don't have them.

swampservers:
I'll merge it, but this should be solveable without having a special check on a specific model

PR #368, opened 2022-07-28
Bug Fixes

Noztik:


lua/cinema/map/sv_mapprops.lua
```diff
@@ -212,3 +212,13 @@ if MAPPROPS_RERUN then
 end
 
 MAPPROPS_RERUN = true
+
+hook.Add("OnEntityCreated","TripminePositionValidator",function(ent,dmginfo) --move this somewhere else
+	if ent:GetClass() == "npc_tripmine" then
+		timer.Simple(0,function()
+			if not util.IsInWorld(ent:GetPos()) then
```
swampservers, line 219: use `ent:TimerSimple(0, function() ...` to ensure the ent is still valid

lua/cinema/map/sv_mapprops.lua
```diff
@@ -212,3 +212,13 @@ if MAPPROPS_RERUN then
 end
 
 MAPPROPS_RERUN = true
+
+hook.Add("OnEntityCreated","TripminePositionValidator",function(ent,dmginfo) --move this somewhere else
```
swampservers: the dmginfo param is not supposed to be there

lua/cinema/weapons/weapon_spraypaint/shared.lua
```diff
@@ -67,6 +67,7 @@ end
 
 function SWEP:PrimaryAttack()
     local ply = self:GetOwner()
+    if not (IsValid(ply) and ply:Alive()) then return end
```
swampservers, line 70: is this to fix an error on server or client? it seems like the weapon shouldn't be able to primary attack unless it has a living owner, unless maybe the localplayer is invalid on client somehow

Noztik: an error is thrown when a person dies while using the spraypaint

swampservers: Oh I see why. The check should be moved down to line 117-ish i think

PR #370, opened 2022-08-07
add support for youtube shorts

bobolobxd:
A revolutionary change to the gmod cinema video playback system. a must have for every server.


swampservers:
ty

PR #376, opened 2022-09-02
fix bug

bobolobxd:
should fix
addons/_swamp/lua/cinema/misc/cl_floatingnames.lua:2: attempt to compare number with nil
when player first connects

### DIFF
--- a/lua/cinema/misc/cl_floatingnames.lua
+++ b/lua/cinema/misc/cl_floatingnames.lua
@@ -92,7 +92,7 @@ local function DrawName(ply, opacityScale)
         end
     end
 
-    if ply.NW.bounty > 0 then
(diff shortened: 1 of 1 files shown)


lua/cinema/misc/cl_floatingnames.lua
```diff
@@ -92,7 +92,7 @@ local function DrawName(ply, opacityScale)
         end
     end
 
-    if ply.NW.bounty > 0 then
+    if ply.NW.bounty and ply.NW.bounty > 0 then
```
swampservers, line 95: can do `if (ply.NW.bounty or 0) > 0 then` as the `or` operator will return the second operand if the first is `nil` (unset) or `false`

PR #379, opened 2022-09-23
Weapon Changes

Metroid55:
Trying to fix the op weapons and get some damage out of the game

### DIFF
--- a/lua/cinema/guns/weapons/gun/shared.lua
+++ b/lua/cinema/guns/weapons/gun/shared.lua
@@ -175,7 +175,7 @@ function GunPerkOverrides(swep, perk)
         t.Damage = swep.Damage * 1.2
         t.CycleTime = swep.CycleTime * 1.1
         t.KickUBase = (swep.KickUBase or 0) * 1.2
-        t.AmmoPriceMod = (swep.AmmoPriceMod or 1) * 1.5
+        t.AmmoPriceMod = (swep.AmmoPriceMod or 1) * 1.4
     elseif perk == "explosive" then
         t.KickUBase = (swep.KickUBase or 0) * 2
     elseif perk == "shothose" then
--- a/lua/cinema/guns/weapons/gun/shared.lua
+++ b/lua/cinema/guns/weapons/gun/shared.lua
@@ -170,13 +170,12 @@ function GunPerkOverrides(swep, perk)
         t.NumPellets = swep.NumPellets * 1.5
         t.PelletSpread = swep.PelletSpread * 1.4
         t.CycleTime = swep.CycleTime * 1.5
-        t.SpawnPriceMod = (swep.SpawnPriceMod or 1) * 1.2
         t.AmmoPriceMod = (swep.AmmoPriceMod or 1) * 2
     elseif perk == "moredamage" then
         t.Damage = swep.Damage * 1.2
         t.CycleTime = swep.CycleTime * 1.1
         t.KickUBase = (swep.KickUBase or 0) * 1.2
-        t.AmmoPriceMod = (swep.AmmoPriceMod or 1) * 1.4
+        t.AmmoPriceMod = (swep.AmmoPriceMod or 1) * 1.5
     elseif perk == "explosive" then
         t.KickUBase = (swep.KickUBase or 0) * 2
     elseif perk == "shothose" then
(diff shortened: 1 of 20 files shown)


lua/cinema/guns/weapons/gun/shared.lua
```diff
@@ -155,20 +155,22 @@ function GunPerkOverrides(swep, perk)
...
-        t.HalfDamageDistance = swep.HalfDamageDistance * 1.7
+        t.HalfDamageDistance = swep.HalfDamageDistance * 1.8
         t.SpawnPriceMod = (swep.SpawnPriceMod or 1) * 1.5
         t.AmmoPriceMod = (swep.AmmoPriceMod or 1) * 1.8
     elseif perk == "selfloading" then
         t.CycleTime = swep.CycleTime * 0.6
         t.UnscopeOnShoot = false
     elseif perk == "boomstick" then
-        t.NumPellets = swep.NumPellets * 2
+        t.NumPellets = swep.NumPellets * 1.5
```
swampservers, line 180: this is too much nerf, since the spread is increased by 1.5x its gonna suck

swampservers:
I'm not sure about all these changes that reduce the base damage by 1, this will result in a lot more almost-kills where someone ends with just a few % health left. Did you test this at all or just putting some values in? I'm fine with nerfing perks like boomstick if they're OP

Metroid55:
The problem is that basically almost every gun is way too cheap and kills too fast. I'm trying to negate some of that, because there is no armor like in css and perks like armor piercing exist to make your weapon more deadly (but no one uses it when the base weapon kills perfectly fine and is cheaper). Most of the killing happens at a close range and weapons like the shotguns, smgs and pistols either 1-shot you (even without perks) or kill you in a split second. I would even go as far as too say that my changes are quite tame and if I could I would increase the price of some weapons/weaponclasses as well (increase pistol prices by 30-40% (more for the deagle and magnum), shotguns by 10%, smgs by 20%, ...)

swampservers:
Maybe we should make it so armor piercing just gives the current default damage values, and non armor piercing divides the damage by what the AP multiplier used to be? That would make all the guns quite a bit weaker. I'm also considering headshot damage and which guns should be a 1 shot headshot.

Metroid55:
Some guns might be a bit too weak with that change (famas for example is already a joke gun + also snipers would feel weird), but we could try that for now and adjust accordingly. You would have to do that though, because I don't know how.

Metroid55:
Updated my pr with the upcoming changes in mind

LegacyValken:
> The problem is that basically almost every gun is way too cheap and kills too fast. I'm trying to negate some of that, because there is no armor like in css and perks like armor piercing exist to make your weapon more deadly (but no one uses it when the base weapon kills perfectly fine and is cheaper). Most of the killing happens at a close range and weapons like the shotguns, smgs and pistols either 1-shot you (even without perks) or kill you in a split second. I would even go as far as too say that my changes are quite tame and if I could I would increase the price of some weapons/weaponclasses as well (increase pistol prices by 30-40% (more for the deagle and magnum), shotguns by 10%, smgs by 20%, ...)

If you want to fix the cheap problem then airsoft weapons need to have their own separate ammo type so players don't just buy airsoft ammo at 100 to 300 points per round when refill cost is meant to be 1.5k to 3k for epics. The cost of constant RDM should eat a bit into profits.

PuraPoint:
> 

Thats a bug, not a feature

swampservers:
So here's the thing. I don't want to merge this because I carefully selected the guns' damage values and dropoff to have a certain number of shots to kill at different ranges. I don't want to lose that.

I think a better way to do this would be to, if the gun does NOT have the armor piercing perk, modify the damage before it is applied so that there is 1 more shot to kill, so to speak.

A quick solution would be something like, on gun/shared.lua line 92, do something like this:
```
local damage = swep.Damage

if swep.GunType == "smg" then
    damage = 100/( (100/damage)+1 )
end
```
then replace all uses of swep.Damage with damage in the function.

This way, all SMGs take 1 more shot to kill.

Metroid55:
Okay I updated it. If you are okay with those changes I can create a new pull request if you want, so that there aren't that many commits.

Metroid55:
Added airsoft fix.

PR #383, opened 2022-10-04
Clean Weapon Changes PR

Metroid55:
Merged everything into one commit to make it cleaner

### DIFF
--- a/lua/cinema/guns/sh_guns.lua
+++ b/lua/cinema/guns/sh_guns.lua
@@ -130,6 +130,17 @@ game.AddAmmoType{
 }
 
 game.AddAmmoType{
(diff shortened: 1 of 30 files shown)


swampservers:
Actually, so, don't do the damage calculation thing in GunPerkOverrides I said before. Instead just modify the damage values in each SWEP, but leave the previous value commented out so its like `SWEP.Damage = 27 --35`

Other than on shotguns, don't just lower the damage by 1. If the damage is greater than 33, subtract 8 from it. Otherwise if the damage is >= 25, subtract 5 from it. This way shots to kill increases by 1.

Metroid55:
Done and ready to be reviewed/merged

WinterPhoenix (approved):

PR #384, opened 2022-10-05
Prevent taunting players from teleporting via /ponyrp.

LegacyValken:


LegacyValken:
Closing, talked to Winter about this problem.

PR #390, opened 2022-10-14
Hopefully the last

Metroid55:
SMGs are the clear winner from the balance changes (which wasn't intended) and will be more in line with the dps where they should be now

-made pistols benefit a bit less from the stat system


Metroid55:
Closing this for now. I'll need to gather more constructive feedback first.

PR #392, opened 2022-10-18
Gun adjustments 

Metroid55:
-added rubber bullets
-smgs and pistols shoot a bit slower (just slightly changes dps, not dmg per mag)
-made smgs slightly more expensive (~200 points more)
-some smgs should be slightly less accurate (can't test this, so we'll see)
-made movement more impactful (added/changed movement modifiers on a lot of the guns)


Metroid55:
Any updates or feedback on this or should I close it?

swampservers:
will work on this uh sometime

Metroid55:
I added a lot more changes to this PR and it would be great to get this accepted and merged. Need to see how it affects RDM and the economy after a while.

PR #395, opened 2023-01-06
Floating friend names are green.

LegacyValken:
Small little change to use steam's in-game activity color to show friends for floating names on players.


swampservers:
The big floating names are gonna be green now? Do you think that's gonna look good?

LegacyValken:
> The big floating names are gonna be green now? Do you think that's gonna look good?

https://i.imgur.com/n4cLq7i.png

I think it looks okay, but I might see your point in that it looks a little funny with the titles being white as well. I don't mind the variety personally and think it's nice to have them stand out a bit, but if you feel it doesn't look good then we can just not do this.

swampservers:
Eh, its not terrible, but I think just leaving it a neutral color is better in general. Maybe you are not actually friends with someone if you need a green name to remind you!

PR #401, opened 2023-07-30
Fix m3u URI parsing

swamproyal:
When we parse M3U, we check if a line contains `.m3u8`. If it does but does not contain `http/s` then we assume it's a partial URI. However we assumed that the URI would be the entire line, which fails in a lot of cases. M3U lets you set the URI with `URI=<URI>`

[examples](https://en.wikipedia.org/wiki/M3U)
```
NAME="English", TYPE=AUDIO, GROUP-ID="audio-stereo-64", LANGUAGE="en", DEFAULT=YES, AUTOSELECT=YES, URI="english.m3u8"
```

failing example from banned.video https://bytehighway.net/3d8d888117f88d252cab8fce4760adbe/manifest/video.m3u8

### DIFF
--- a/lua/cinema/theater/services/sh_hls.lua
+++ b/lua/cinema/theater/services/sh_hls.lua
@@ -28,7 +28,7 @@ if CLIENT then
     end
 
     function SERVICE:LoadVideo(Video, panel)
-        panel:EnsureURL("https://swamp.sv/s/cinema/file.html")
+        panel:EnsureURL("https://swamp.sv/s/cinema/hls.html")
 
         panel:AddFunction("gmod", "loaded", function()
             self:SeekTo(CurTime() - Video:StartTime(), panel)
--- a/lua/cinema/theater/services/sv_hls.lua
+++ b/lua/cinema/theater/services/sv_hls.lua
@@ -34,7 +34,7 @@ sv_GetVideoInfo.hls = function(self, key, ply, onSuccess, onFailure)
             if string.find(line, ".m3u8") then
                 local streamurl
             
-                if string.find(line, "http.://") then
+                if string.find(line, "https?://") then
                     streamurl = line
                 else -- relative path
                     local uri = string.match(line, 'URI="(.-)"')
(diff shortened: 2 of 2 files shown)


WinterPhoenix (changes requested):
Please tweak the string.find directly above and use this pattern instead. `.` is far too greedy for that:
```
string.match(streamurl, "https?://")
```

swamproyal:
> Please tweak the string.find directly above and use this pattern instead. `.` is far too greedy for that:
> 
> ```
> string.match(streamurl, "https?://")
> ```

I don't think this will work. We're basically checking each line of the m3u8 file for if it contains `http/s` currently. This fails if we have a URI which is relative and does not contain `http/s`. I added a commit to clarify the logic a little bit

WinterPhoenix:
Sorry, I mean `string.find(line, "https?://")`. The `.` character means it'll accept *any* character there, not just `s`. Also it means there *must* be a character there, so that code won't work for URLs beginning in `http://`. The `?` character means 0 or 1 matches

swamproyal:
> Sorry, I mean `string.find(line, "https?://")`. The `.` character means it'll accept _any_ character there, not just `s`. Also it means there _must_ be a character there, so that code won't work for URLs beginning in `http://`. The `?` character means 0 or 1 matches

Sounds good. I misunderstood what you wanted the first time. This makes total sense. I've commited that change

lua/cinema/theater/services/sh_hls.lua
```diff
@@ -28,7 +28,7 @@ if CLIENT then
     end
 
     function SERVICE:LoadVideo(Video, panel)
-        panel:EnsureURL("https://swamp.sv/s/cinema/file.html")
+        panel:EnsureURL("https://swamp.sv/s/cinema/hls.html")
```
swamproyal, line 31: I realized, for some reason, we weren't using hls.html for m3u8 playlists. I have no idea why this changed but yeah this fixes some instances. I don't think file.html handles playlist chunks or something because it starts at the beginning in some cases even though it will play the video.

WinterPhoenix: file.html uses Video.JS while hls.html uses hls.js, so that would make sense

WinterPhoenix (approved):

PR #402, opened 2023-07-30
Add kick.com support

swamproyal:
I tested to the best of my ability but I dont have a way to run the sv code

tldr:
- Get the key from the url eg. `kitty` from `https://kick.com/kitty`
- Server hits kick API at `https://kick.com/api/v2/channels/$key` example: https://kick.com/api/v2/channels/kitty
- We get the thumbnail and stream title from that json
- Load the video with the kick player at `https://player.kick.com/$key`
- After we load the kick player, we need to press the unmute button on the page automatically because it always starts muted

SetVolume just finds all the video players on the page and sets the element volume for each

### DIFF
--- a/lua/cinema/theater/services/sh_kick.lua
+++ b/lua/cinema/theater/services/sh_kick.lua
@@ -1,0 +1,32 @@
+﻿-- This file is subject to copyright - contact swampservers@gmail.com for more information.
+local SERVICE = {}
+SERVICE.Name = "Kick"
+SERVICE.NeedsCodecs = true
+SERVICE.NeedsChromium = true
+
+function SERVICE:GetKey(url)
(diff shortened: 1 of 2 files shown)


WinterPhoenix:
Are you sure the video always starts muted, or is it just that the browser you're testing with has autoplay (with sound) disabled?

swamproyal:
> Are you sure the video always starts muted, or is it just that the browser you're testing with has autoplay (with sound) disabled?

Yeah. The behavior in game is that it starts muted.

swamproyal:
bump, could we work to get this in this weekend?

WinterPhoenix (approved):
Looks good. We may need to adjust API calls in the future, as it's currently undocumented, but this works atm.

Also, it'd be awesome to see VOD and Clip support in the future

PR #405, opened 2023-08-31
Merge culling optimizations for better swamp performance

insertname1180:
This would be perhaps a neat addition to swamp.

Not sure behind the licences behind this though.. I would need help on enlightening on this part of the subject because I do see a need for this on swamp servers. I would like input if swampservers, WinterPhoenix or Kefta have any ideas on how to implement this mod onto the server.

I am the creator of the mod

If it's too buggy, please let me know and I'll be happy to adjust the mod to swamp cinema's needs.

I thought it'd be cool to have this mod on swamp as an implementation and will bend the mod to meet the needs of swampservers.

Please message back if possible and I will adjust the mod accordingly

Let me know these following things:

- Desired field of view
- If you would like the option of culling behind walls (when in players field of view) ( I don't know if this will bug the minecraft mines that swamp has )
- Update rate

Also, there can be certain entity types (any type you want) that can be excluded from the mod.

The mod is on my github and on the workshop.

### DIFF
--- a/lua/autorun/client/TrueCulling.lua
+++ b/lua/autorun/client/TrueCulling.lua
@@ -1,0 +1,57 @@
+jit.on()
+
+CULLFRONT = math.sin(0.57357644) -- Degrees forward angle in sine
+
+function main()
+	allEnts = ents.GetAll()
+	if engine.TickCount()%4 == 0 then
+		for a, ents_ in pairs(allEnts) do
+			if engine.TickCount()%10 == 0 then
+				ents_:SetNoDraw(true)
+			end
+			if engine.TickCount()%10 == 0 then
+				if ents_:GetNoDraw(true) then
+					if util.IsPointInCone(ents_:GetPos(), LocalPlayer():GetPos(), LocalPlayer():GetAimVector(), CULLFRONT, 100000) then
+						if LocalPlayer():IsLineOfSightClear(ents_:GetPos() + Vector(-25,-25,-25)) or LocalPlayer():IsLineOfSightClear(ents_:GetPos() + Vector(10,10,10)) then
+							if ents_:GetNoDraw(true) then 
+								ents_:SetNoDraw(false)
+							end
+						end
+					end
+				end
+			else
+				return
+			end
+			if engine.TickCount()%10 == 0 then
+				if ents_:IsPlayer() or ents_:GetClass() == "viewmodel" then
+					if ents_:GetNoDraw(true) then            
+						ents_:SetNoDraw(false)
+					end
+				end
+			else
+				return
+			end
+			if engine.TickCount()%10 == 0 then
+				if ents_:GetPos():Distance(LocalPlayer():GetPos()) < 100 then
+					if ents_:GetNoDraw(true) then				
+						ents_:SetNoDraw(false)
+					end			
+				end
+			else
+				return
+			end
+			if engine.TickCount()%10 == 0 then
+				if ents_:GetClass() == "10C_BaseFlex" then
+					ents_:SetNoDraw(true)
+				end
+			else
+				return
+			end
+		end
+	else
+		return
+	end
+end
+
+hook.Add("Think", "cullLoop", main)
+hook.Add("InitPostEntity", "initents", main)
--- a/lua/autorun/client/TrueCulling.lua
+++ b/lua/autorun/client/TrueCulling.lua
@@ -40,12 +40,23 @@ function main()
 			else
 				return
 			end
-			if engine.TickCount()%10 == 0 then
-				if ents_:GetClass() == "10C_BaseFlex" then
-					ents_:SetNoDraw(true)
+			if system.IsLinux() or system.IsOSX() then
+				if engine.TickCount()%10 == 0 then
+					if ents_:GetClass() == "10C_BaseFlex" then
+						ents_:SetNoDraw(true)
+					end
+				else
+					return
 				end
-			else
-				return
+			end
+			if system.IsWindows() then
+				if engine.TickCount()%10 == 0 then
+					if ents_:GetClass() == "class C_BaseFlex" then
+						ents_:SetNoDraw(true)
+					end
+				else
+					return
+				end
 			end
 		end
 	else


lua/autorun/client/TrueCulling.lua
```diff
@@ -0,0 +1,68 @@
+jit.on()
+
+CULLFRONT = math.sin(0.57357644) -- Degrees forward angle in sine
+
+function main()
+	allEnts = ents.GetAll()
```
swampservers, line 6: So, you're calling ents.GetAll, a very expensive function, every single frame...

lua/autorun/client/TrueCulling.lua
```diff
@@ -0,0 +1,68 @@
...
+	if engine.TickCount()%4 == 0 then
```
swampservers, line 7: And yet, you are only using the result of that call, every forth frame?

swampservers:
On initial glance I can tell this would cause far more damage to performance than you would get out of it. I can help you write this properly though.

Why do you say you are not sure about the licenses behind it if you are the creator of the mod...? If you made it, you determine the license (although by submitting it here you are giving us the right to use it under our CLA on the main page)

lua/autorun/client/TrueCulling.lua
```diff
@@ -0,0 +1,68 @@
...
+		for a, ents_ in pairs(allEnts) do
+			if engine.TickCount()%10 == 0 then
```
swampservers, line 9: So this "performance optimization" addon computes the static value `engine.TickCount()%10 == 0` 5 times for each entity? Do you know what a variable is?

lua/autorun/client/TrueCulling.lua
```diff
@@ -0,0 +1,68 @@
...
+				return
+			end
+			if engine.TickCount()%10 == 0 then
+				if ents_:IsPlayer() or ents_:GetClass() == "viewmodel" then
+					if ents_:GetNoDraw(true) then            
+						ents_:SetNoDraw(false)
+					end
+				end
+			else
+				return
```
swampservers, line 32: Isn't this supposed to check/update nodraw on each entity? Once one of these returns is hit, it will break out.

lua/autorun/client/TrueCulling.lua
swampservers: You should be using the hooks EntityCreated/NetworkEntityCreated/EntityRemoved (if I remember those names correctly) to maintain a single global table of all entities. And the entities should be keys so you are not doing an O(N) table.remove when an entity is deleted.

lua/autorun/client/TrueCulling.lua
```diff
@@ -0,0 +1,68 @@
...
+				end
+			end
+		end
+	else
+		return
+	end
+end
+
+hook.Add("Think", "cullLoop", main)
+hook.Add("InitPostEntity", "initents", main)
```
swampservers, line 68: I don't see the need to call this on init when you are calling it every frame anyway

lua/autorun/client/TrueCulling.lua
swampservers: Also you need to be using local variables so it is not looking up the variable name in the global namespace each access

lua/autorun/client/TrueCulling.lua
```diff
@@ -0,0 +1,68 @@
...
+			if system.IsLinux() or system.IsOSX() then
+				if engine.TickCount()%10 == 0 then
+					if ents_:GetClass() == "10C_BaseFlex" then
+						ents_:SetNoDraw(true)
+					end
+				else
+					return
+				end
+			end
+			if system.IsWindows() then
```
swampservers, line 52: should do `if windows then (code for windows) else (code for linux/osx) end`

lua/autorun/client/TrueCulling.lua
```diff
@@ -0,0 +1,68 @@
+jit.on()
+
+CULLFRONT = math.sin(0.57357644) -- Degrees forward angle in sine
```
swampservers, line 3: Is this supposed to be related to the field of view? FOV is not static, and if you want to do this properly, it's not that hard to check the actual frustum instead of bounding it with a cone

lua/autorun/client/TrueCulling.lua
```diff
@@ -0,0 +1,68 @@
...
+	allEnts = ents.GetAll()
+	if engine.TickCount()%4 == 0 then
+		for a, ents_ in pairs(allEnts) do
+			if engine.TickCount()%10 == 0 then
+				ents_:SetNoDraw(true)
+			end
+			if engine.TickCount()%10 == 0 then
+				if ents_:GetNoDraw(true) then
+					if util.IsPointInCone(ents_:GetPos(), LocalPlayer():GetPos(), LocalPlayer():GetAimVector(), CULLFRONT, 100000) then
+						if LocalPlayer():IsLineOfSightClear(ents_:GetPos() + Vector(-25,-25,-25)) or LocalPlayer():IsLineOfSightClear(ents_:GetPos() + Vector(10,10,10)) then
```
swampservers, line 15: Can you explain what this is supposed to do? It seems like you are checking tracing visibility to two arbitrary points near the object with no regard to the size of that object.

lua/autorun/client/TrueCulling.lua
swampservers: If you want to run this every Nth frame for speed because it doesn't need to be updated every frame, this will technically result in microstutter where every Nth frame is slower. What you can do instead is amortize the checks, something like this:

```
for ent,_ in pairs(ents) do
    if ent:EntIndex() % N == FrameNumber() % N then
        (do checks)
    end
end
```

lua/autorun/client/TrueCulling.lua
swampservers: I used to have a script that did what I think you are trying to do here, where it would choose a random point within the object bounding box, and it would only hide the object if it failed all of the past 10 or so checks. That way it would basically stochastically estimate the visibility of the entire object. I think I later removed it because it made little difference and wasn't worth the cost of checking the visibility.

lua/autorun/client/TrueCulling.lua
```diff
@@ -0,0 +1,68 @@
...
+						ents_:SetNoDraw(false)
+					end			
+				end
+			else
+				return
+			end
+			if system.IsLinux() or system.IsOSX() then
+				if engine.TickCount()%10 == 0 then
+					if ents_:GetClass() == "10C_BaseFlex" then
+						ents_:SetNoDraw(true)
```
swampservers, line 46: So you're always hiding these BaseFlex entities. Why?

lua/autorun/client/TrueCulling.lua
```diff
@@ -0,0 +1,68 @@
...
+							end
+						end
+					end
+				end
+			else
+				return
+			end
+			if engine.TickCount()%10 == 0 then
+				if ents_:IsPlayer() or ents_:GetClass() == "viewmodel" then
+					if ents_:GetNoDraw(true) then
```
swampservers, line 27: It seems like many of these checks can be removed by structuring the code properly

lua/autorun/client/TrueCulling.lua
swampservers: I have some code somewhere that will get the actual view frustum after other scripts have modified it, I can dig it out for you if you want to try to fix this

lua/autorun/client/TrueCulling.lua
swampservers: Also you previously already checked `engine.TickCount()%4==0`. So is this supposed to run every forth tick, or every tenth? Also you should use FrameNumber() because Think runs every frame (client), not every tick (server).

insertname1180:
I do not agree to the CLA

insertname1180:
Thank you swampservers for teaching me a better way to do it

insertname1180:
Thank you to swampservers for the sage advice

Kefta:
Solved in private discussion.

PR #406, opened 2023-09-05
Add adjustable weapon volume

swamproyal:
This allows you to set weapon related sound volumes via a convar float. This is a pretty popular request on the server where you like to hear player sounds and you're in a theater or whatever and the weapons get annoying. Could probably also be added to the tab menu settings.


lua/cinema/misc/cl_weaponvolume.lua
```diff
@@ -0,0 +1,23 @@
...
+local weapon_sound_patterns = {
+    "weapon",
+    "bullet",
+    "garand",
+    "throatneck",
+    "isissong"
+}
+
+hook.Add("EntityEmitSound", "EntityEmitSound_WeaponVolume", function(data)
+    for k, v in pairs(data) do
```
Kefta: Remove the loop and access the SoundName key directly like ``local sound = data.SoundName``

lua/cinema/misc/cl_weaponvolume.lua
```diff
@@ -0,0 +1,23 @@
+-- This file is subject to copyright - contact swampservers@gmail.com for more information.
+local swamp_weaponvolume = CreateClientConVar("swamp_weaponvolume", "1", true, false, "", 0.0, 1.0)
+
+local weapon_sound_patterns = {
```
Kefta, line 4: What are the exact file paths that you are trying to capture with these patterns? This will probably catch some false positives without a stricter pattern.

lua/cinema/misc/cl_weaponvolume.lua
```diff
@@ -0,0 +1,23 @@
...
+    "throatneck",
+    "isissong"
+}
+
+hook.Add("EntityEmitSound", "EntityEmitSound_WeaponVolume", function(data)
+    for k, v in pairs(data) do
+        if (k == "SoundName") then
+            for i = 1, #weapon_sound_patterns do
+                if string.find(v, weapon_sound_patterns[i]) then
+                    data.Volume = swamp_weaponvolume:GetFloat()
```
Kefta: This should be scaled by the previously defined Volume (if there is one) in case the weapon specifies its own volume in EmitSound. ``data.Volume = swamp_weaponvolume:GetFloat() * (data.Volume or 1)``

Kefta (changes requested):

lua/cinema/misc/cl_weaponvolume.lua
Kefta: I see you beat me to it

lua/cinema/misc/cl_weaponvolume.lua
swamproyal: Any sounds emitted by weapons which cause damage to the player. I'm not sure a stricter pattern is necessary, tbh.

lua/cinema/misc/cl_weaponvolume.lua
swamproyal: Thanks, good catch

lua/cinema/misc/cl_weaponvolume.lua
Kefta: Stricter patterns would prevent any possible chance of false positives passing through if any of these names are used in part of the directory or file name for non-weapons + be slightly more efficient with how patterns are implemented.

A better implementation might be to skip the patterns and instead just check if the entity emitting the sound is a weapon, making it future-proof as well. There might be cases in Swamp SWEPs where the sound is emitted from the player instead, but either a hack can be added to still check for the patterns + the player's active weapon, or I can go in and fix those cases.

swamproyal: So I think checking if the entity emitting the sound is a weapon isn't so desirable because a lot of the SWEPs are weapons if I'm not mistaken - such as the "free toys". This feature should not affect those sounds. Like I mentioned, I think this should only affect weapons which cause damage to a player.

swamproyal: Side note: I thought about branding this as "adjustable gun sounds" but I wanted it to capture more SWEPs than just guns, such as jihad bomb, throatneck slitter, big bomb, etc...

Kefta: In that case, though it sounds like more complicated option, I think it would be best to insert something like SWEP.IsSwampDamageWeapon = true into all the SWEPs that should be affected, then the hook would check for that key in the entity that emitted the sound. This would prevent targeting any unintentional sounds, make it easily expandable for SWEP authors without having to remember to go into this file, and be far more efficient (not that this is currently a frame killer or anything). Guns should be easy to apply this property to since they all resolve to their own common weapon bases, but other than the weapons you just listed, what else should receive this property?

Separately, a convar could be added to change the volume of all weapon sounds regardless of their type but I don't know if you wanted to not affect those sounds for a reason.

swamproyal: Probably a better option here is to effectively have a whitelist on the toys*. I think these are the only SWEPs which should remain unaffected by the weapon volume. It would be annoying to enumerate all of the weapons which need to be affected because that includes all of the CS stuff, HL2 stuff, a mix of custom swamp stuff, probably other stuff. Would you be content with me adding a property to the toys e.g. `SWEP.IsToy = true`?

~~To clarify, should we get the Entity from [EntityEmitSound](https://wiki.facepunch.com/gmod/GM:EntityEmitSound) table, get the active weapon, i.e. `Entity:GetActiveWeapon()`, then check if `IsToy` exists? If so, apply the volume scaling?~~

How do you recommend getting the entity which triggered the sound from the EntityEmitSound callback table?

*Specifically, the toys are these if you aren't familar: https://imgur.com/H37JZWW.png

Kefta: Good idea, could you call it something like``SWEP.IsPassiveWeapon`` (specifically with the Weapon suffix since IsPassive could conflict with some weapon base's "IsPassive" method for checking weapon lowering) to notate that it should apply to anything that doesn't shoot, blow up, or otherwise do damage? I would imagine the Peacekeeper (and the taser and pickaxe?) should also be affected by this convar despite being a toy(s).

I forgot the weapons were public so you can change this if need-be, but the idea is that the entity should be used directly and no hacks would be needed. If any SWEP is running EmitSound on the player instead of the weapon, it needs to be corrected in code. I see that pretty much everything uses ExtEmitSound so that probably won't be a problem.

Kefta: Actually, there are some valid cases in which a damaging weapon will create sounds on a non-weapon entity (taser, crusader's sword) and some incorrect ones that do need to be fixed (the garand). However, due to this, the solution of checking a weapon property alone won't work. I need to do some brainstorming but it might require solidifying usage of the ExtEmitSound function or overriding EmitSound in the metatable.

swamproyal: >then the hook would check for that key in the entity that emitted the sound. 

How would this be possible?

EntityEmitSound callback data table contains an Entity property which represents the Entity that generated the sound. In the case of a gun shot sound, that entity is the Player. There isn't a way to know if the sound event was generated from a SWEP/Weapon. This would only work for actual weapon entities, e.g. bomb, mines, etc...

swamproyal: Bump @Kefta, any thoughts on my comment? I'm not sure that we can only use the entity in EntityEmitSound callback data table without looking at the sound file.

Kefta: Still thinking, I'm looking further to a swamp-specific EmitSound override that would replace both ExtEmitSound and provide more info to this hook

swamproyal: Any further thoughts? People would like to start using this. We could just go back to my original implementation with the regex patterns and just make them more specific.

swamproyal:
Bump @Kefta

WinterPhoenix (commented):
So, I think this should be added to the Scoreboard UI before it's merged, along with the other volume sliders. Beyond that it's probably good to go...although EntityEmitSound is generally unreliable so YMMV

swampservers:
This should probably be implemented alongside the main clientside volume stuff to make sure it doesn't conflict.

Also: There are too many volume sliders. This and the loading volume should be in a collapsable/popout or something.

swamproyal:
> So, I think this should be added to the Scoreboard UI before it's merged, along with the other volume sliders. Beyond that it's probably good to go...although EntityEmitSound is generally unreliable so YMMV

I tested it on the server again and it works still. I don't think I have access to the scoreboard code so I can't help with that.

WinterPhoenix:
Of course it's in restricted for some reason...

Edit - It's here now: https://github.com/swampservers/contrib/blob/master/lua/cinema/cl_ui/settings.lua

swamproyal:
@WinterPhoenix I added a scoreboard slider for the weapon volume. If you think that is too many sliders on the scoreboard menu, I think we should add this as a convar only setting regardless.

PR #407, opened 2023-09-05
Change some EmitSound targets from the player to the weapon

Kefta:
As part of the discussion in https://github.com/swampservers/contrib/pull/406.


WinterPhoenix (approved):
lgtm

PR #408, opened 2023-09-10
Revert to using file.html for .m3u8

swamproyal:
video.js seems much more reliable for seeking. There have been problems the last week or so where videos take forever to load or have problems when seeking. I confirmed this in the browser by just using hls.html directly. file.html didn't have these problems. So, we should revert.

I'll have to look into what's going on here. I would like to use hls.js because video.js doesn't work with all m3u8 playlist formats. Maybe it needs an update?

### DIFF
--- a/lua/cinema/theater/services/sh_hls.lua
+++ b/lua/cinema/theater/services/sh_hls.lua
@@ -28,7 +28,7 @@ if CLIENT then
     end
 
     function SERVICE:LoadVideo(Video, panel)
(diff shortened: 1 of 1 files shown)


swamproyal:
bump, this is important

WinterPhoenix (approved):
Sure let's give it a shot. This will probably require further investigation though yeah

swamproyal:
> Sure let's give it a shot. This will probably require further investigation though yeah

Okay, I would appreciate if important fixes could get looked at sooner. I'll be looking into this one more.

PR #410, opened 2023-09-16
Fix starting volume for kick videos

swamproyal:


WinterPhoenix (changes requested):
This isn't reliable. Any random video element on the page could be an ad or something. And why are you including `<audio>` as part of the search?

Find the video element for the player more specifically. If it's got a specific id, use that and grab it directly. If not, maybe a parent element has an id, and you look inside of it.

swamproyal:
> This isn't reliable. Any random video element on the page could be an ad or something. And why are you including `<audio>` as part of the search?
> 
> Find the video element for the player more specifically. If it's got a specific id, use that and grab it directly. If not, maybe a parent element has an id, and you look inside of it.

What do you mean with regards to reliability? Did you test it? I iterate through anything that could be making sound on the webpage. That seems like the most reliable way to do it. What is a sitation where that isn't the desired behavior? Is there a reason to find only the kick video player? My SetVolume for kick does literally the exact same thing.

swamproyal:
I thought about it more and I see what could go wrong now - you don't want an ad or something to call determine the "readiness." I'll fix it by checking for the kick player element I'd.

I thought you were referring to the reliability of setting the volume to 0 initially.

swamproyal:
I added a check to ensure we know the actual kick player is loaded

WinterPhoenix (approved):
Works for me

PR #411, opened 2023-09-16
Add rumble.com support

swamproyal:
I can't test the server code but I think it will work.


WinterPhoenix:
What's the motivation for this while we already have Twitch and Kick? I certainly haven't heard of this site

swamproyal:
> What's the motivation for this while we already have Twitch and Kick? I certainly haven't heard of this site

Many who were deplatformed by Google et all are now here. We use this site a lot. It isn't a small site, at this point it is pretty well established.

WinterPhoenix (approved):

PR #412, opened 2023-09-17
Add Infowars video service

swamproyal:
Don't have the ability to fully test but I think it will work


swamproyal:
Yes, we do actually use this site on the server and have to manually find the m3u8 playlist link. It would be great if it was natively supported.

swamproyal:
I've pushed some improvements

WinterPhoenix (approved):

PR #413, opened 2023-09-24
Add CORS proxy for file and hls

swamproyal:
Certain popular websites on the server have had a CORS policy change, breaking requests. I have a CORS proxy as a work around. It can handle high bandwidth.

### DIFF
--- a/lua/cinema/theater/services/sh_file.lua
+++ b/lua/cinema/theater/services/sh_file.lua
@@ -155,7 +155,17 @@ if CLIENT then
                 }, 100);
             ]])
         else
-            panel:QueueJavascript(string.format("th_video('%s');", string.JavascriptSafe(Video:Key())))
+            panel:QueueJavascript([[
+                th_video(']] .. string.JavascriptSafe(Video:Key()) .. [[');
+                function checkLoaded() {
+                    if (player.error().code === 4) {
+                        // Try a CORS proxy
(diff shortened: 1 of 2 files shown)


WinterPhoenix:
There's...no way we're gonna proxy stuff through a web server we don't control lol

If this proves to be a recurring issue, make your case for it, and we can implement something like this on swamp.sv

swamproyal:
> There's...no way we're gonna proxy stuff through a web server we don't control lol
> 
> If this proves to be a recurring issue, make your case for it, and we can implement something like this on swamp.sv

I mean, you don't control any of the web servers supported on cinema. I think @swampservers would have a differing opinion for swamp.sv proxying requests because it means swamp is now implicated should there be some issue come up. Whereas here, a nonaffiliated service is handling everything. This has been done before in the code for dlive support, I believe.

If you look at recent requests on the server, or play the server, a large majority of the requests in the Movie Theater are using this proxy now. Players have to do this manually and it is difficult.

WinterPhoenix:
There is a difference between players explicitly using a third-party site or proxy and it being baked into our code. From multiple perspectives, including privacy, security, and liability

swamproyal:
> There is a difference between players explicitly using a third-party site or proxy and it being baked into our code. From multiple perspectives, including privacy, security, and liability

I understand this which is, why I'm not going to push for it and find a better solution.

PR #416, opened 2023-10-10
Fix rumble seek

swamproyal:


swamproyal:
@WinterPhoenix could you accept this one asap

swamproyal:
Also, can you add rumble to the video request browser home page?

WinterPhoenix:
querySelectorAll is slow. Just find the video player and persist it as a var; no need to scan the entire page every time the volume changes or the video needs to be seeked

swamproyal:
> querySelectorAll is slow. Just find the video player and persist it as a var; no need to scan the entire page every time the volume changes or the video needs to be seeked

I've simplified it and addressed this concern.

swamproyal:
@WinterPhoenix does it look good to you?

WinterPhoenix (approved):

WinterPhoenix:
There you go. lgtm

PR #420, opened 2023-10-14
Remove invalid bumpmap in maze model texture

swamproyal:
This fixes the maze wall texture on Linux

### DIFF
--- a/materials/models/swamponions/maze.vmt
+++ b/materials/models/swamponions/maze.vmt
@@ -1,7 +1,6 @@
 "VertexLitGeneric"
 {
-	"$basetexture"	"egypt\flat_wall_02"
-	"$bumpmap"	"egypt\flat_wall_normal"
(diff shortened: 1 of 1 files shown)


WinterPhoenix:
I think we may be changing this in Map v4. I'll wait for @PYROTEKNIK's input

PYROTEKNIK:
> This fixes the maze wall texture on Linux

what is the issue on linux?

swamproyal:
> > This fixes the maze wall texture on Linux
> 
> what is the issue on linux?

The texture doesn't work. It's black.

swamproyal:
It's not a new issue either, it's been this way since the v3 release in 2020

PYROTEKNIK:
I think we should wait until the next map version to look at this. The upcoming version uses lightmappedgeneric with fake lightmaps

WinterPhoenix:
Not gonna touch it in that case. v4 is close

PR #423, opened 2023-10-14
Fix pony eye texture on Linux

swamproyal:


Kefta (approved):

WinterPhoenix:
oh my god is that why?

PR #430, opened 2023-12-19
Fix kick API

swamproyal:
They check for js when using their api now. We can just do it clientside

### DIFF
--- a/lua/cinema/theater/services/sh_kick.lua
+++ b/lua/cinema/theater/services/sh_kick.lua
@@ -16,6 +16,46 @@ function SERVICE:GetKey(url)
 end
 
 if CLIENT then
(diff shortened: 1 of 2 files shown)


WinterPhoenix:
Can't reproduce it failing. Is there a specific stream it's not working with or something?

Or do you just mean the Cloudflare CAPTCHA is showing up for you?

swamproyal:
Hmm, there was a check a couple weeks ago when I made this that ensured js was running in the browser. Otherwise, it would 403 if I remember. I think this change is better anyway since it aligns with the pattern we've been switching to requesting everything clientside.

Also, we don't want to hit some limit or get blocked serverside

WinterPhoenix:
I'll merge it if it becomes a problem. We don't really wanna rely on clientside stuff more than we have to

PR #431, opened 2023-12-20
Fix bad tag condition

swamproyal:
I fixed a condition where you're in a protected theater sitting down and can still get tagged (and get blown up after a timer expires).


Metroid55 (commented):
Shouldn't it be IsProtected instead of InTheater? This would just make it so you can't tag anyone when they are sitting down in theaters regardless of protection status.

swamproyal:
> Shouldn't it be IsProtected instead of InTheater? This would just make it so you can't tag anyone when they are sitting down in theaters regardless of protection status.

Yep

WinterPhoenix (approved):

PR #433, opened 2024-01-11
Thirdperson changes

swamproyal:
- Removed the thirdperson "sideview" mode per request of players (people think its dumb/annoying rather than useful)
- Reduced the interpolation when switching from firstperson to thirdperson so it feels more instant
- fixed player being invisible when taunting
- other code cleanup

### DIFF
--- a/lua/cinema/misc/cl_calcview.lua
+++ b/lua/cinema/misc/cl_calcview.lua
@@ -10,7 +10,6 @@ timer.Simple(0, function()
     end
 end)
 
-local thirdperson_enabled = false
 local thirdperson_lerp = 0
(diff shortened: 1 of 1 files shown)


WinterPhoenix (changes requested):
There's code elsewhere that uses the THIRDPERSON variable. Namely the chat command, weapon_taser, and GPoser

swamproyal:
> There's code elsewhere that uses the THIRDPERSON variable. Namely the chat command, weapon_taser, and GPoser

Glad you noticed that. I've fixed it

WinterPhoenix (approved):

PR #436, opened 2024-01-23
Fix debug locations missing font

swamproyal:


WinterPhoenix:
This code'll be replaced when Map v4 merges, which will be very soon, so in the interest of avoiding merge conflicts I'm gonna decline this.

PR #446, opened 2024-03-16
Fix bflix/Nites

swamproyal:
They moved to a new host, I guess we should still call the service bflix since thats what they use internally for domains and stuff?

### DIFF
--- a/lua/cinema/theater/services/sh_bflix.lua
+++ b/lua/cinema/theater/services/sh_bflix.lua
@@ -86,7 +86,7 @@ if CLIENT then
                 -- The main video page
                 if url == key then
                     self:QueueJavascript([[
-                        const title = document.querySelector('body div#aa-wp div.bd.cont section.section.single div.rw aside.left.cl1 article.post.single div.dfxb div header.entry-header h1.entry-title').textContent;
+                        const title = document.querySelector('h1[itemprop="name"]').textContent;
                         const thumbnailUrl = document.querySelector('[itemprop=thumbnailUrl]').content;
                         exTheater.onVideoInfoReady({
                             "title": title,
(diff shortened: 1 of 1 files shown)


lua/cinema/theater/services/sh_bflix.lua
```diff
@@ -81,27 +83,34 @@ if CLIENT then
                     end
                 end)
 
+                -- The main video page
                 if url == key then
+                    self:QueueJavascript([[
+                        const title = document.querySelector('body div#aa-wp div.bd.cont section.section.single div.rw aside.left.cl1 article.post.single div.dfxb div header.entry-header h1.entry-title').textContent;
```
WinterPhoenix: this is way too specific. will almost certainly break

lua/cinema/theater/services/sh_bflix.lua
```diff
@@ -119,28 +128,19 @@ if CLIENT then
     end
 
     function SERVICE:LoadVideo(Video, panel)
-        panel:EnsureURL(Video:Key())
+        panel:EnsureURL(Video:Data())
```
WinterPhoenix: You should be checking if Data ~= "" here. Data is just an empty string, not the Key, by default

WinterPhoenix (changes requested):

lua/cinema/theater/services/sh_bflix.lua
swamproyal: done

lua/cinema/theater/services/sh_bflix.lua
swamproyal: I made it better/hopefully more robust

WinterPhoenix (approved):

PR #447, opened 2024-03-20
Allow optional trailing / for bflix urls

swamproyal:
Their webserver often enforces trailing / so we need this


swamproyal:
Disregard. I have other changes I need to make now

PR #448, opened 2024-03-20
Bflix/nites fixes

swamproyal:
Bflix original site has returned. I added back the original bflix service that was working before. I created new files for the new nites site.

I also had to allow trailing slashes for nites because their webserver always adds a trailing slash.


WinterPhoenix:
If it's the same service with the same code, just make one service that matches both domain names

swamproyal:
> If it's the same service with the same code, just make one service that matches both domain names

I don't think we should consider it the same video service. Let's consider them completely separate.

WinterPhoenix (approved):

PR #452, opened 2024-04-27
Fix some location related issues

Metroid55:
Fixes not being able to build there

### DIFF
--- a/lua/cinema/construction/sh_trash.lua
+++ b/lua/cinema/construction/sh_trash.lua
@@ -90,7 +90,6 @@ TrashLocationOverrides = {
     ["SportZone"] = TRASHLOC_BUILD,
     ["SushiTheater Attic"] = TRASHLOC_BUILD,
     ["SushiTheater Basement"] = TRASHLOC_BUILD,
-    ["SushiTheater Third Floor"] = TRASHLOC_NOSPAWN,
     ["SushiTheater Second Floor"] = TRASHLOC_BUILD,
     ["SushiTheater First Floor"] = TRASHLOC_BUILD,
     ["The Pit"] = TRASHLOC_BUILD,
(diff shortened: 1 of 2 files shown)


lua/cinema/construction/sh_trash.lua
```diff
@@ -90,7 +90,6 @@ TrashLocationOverrides = {
     ["SportZone"] = TRASHLOC_BUILD,
     ["SushiTheater Attic"] = TRASHLOC_BUILD,
     ["SushiTheater Basement"] = TRASHLOC_BUILD,
-    ["SushiTheater Third Floor"] = TRASHLOC_NOSPAWN,
```
WinterPhoenix, line 93: Removing the override won't allow you to build there. The default is NOSPAWN if it's a Theater (and you don't own it) or NOBUILD otherwise:
https://github.com/swampservers/contrib/blob/7e25bd39043212f480a56ab73764c823734e0828/lua/cinema/construction/entities/prop_trash/shared.lua#L298-L310

You want to set it to TRASHLOC_BUILD if you want to allow building in a location.

WinterPhoenix (changes requested):

lua/cinema/construction/sh_trash.lua
Metroid55: No this is correct, because I want it to function like a private theater where only the owner can build and not like the auditorium where everyone can build. What you posted gets overridden later when you are the owner. (Look at line 386)
https://github.com/swampservers/contrib/blob/7e25bd39043212f480a56ab73764c823734e0828/lua/cinema/construction/entities/prop_trash/shared.lua#L365-L391

Metroid55: Actually I think this is more relevant. The code above is just allowing you to tape, but this should allow the prop to exist in your theater.
https://github.com/swampservers/contrib/blob/7e25bd39043212f480a56ab73764c823734e0828/lua/cinema/construction/entities/prop_trash/shared.lua#L346-L354

WinterPhoenix: Okay, I see what you're going for. Removing that override will in fact make SushiTheater Third Floor act like a normal private theater

WinterPhoenix (approved):

PR #458, opened 2024-07-27
Rumble fixes

swamproyal:
- No more ads
- Fixed live streams
- Fixed videos getting out of sync


WinterPhoenix (approved):
I'm not sure this will work with CORS, but I'll accept it so you can test it

swamproyal:
> I'm not sure this will work with CORS, but I'll accept it so you can test it

Yeah, i've been testing it live on the server and it's been fine. If it breaks or has issues I'll fix it.

PR #463, opened 2024-10-20
Add monster can open sound effect for reload, deploy, & init

Narshalbarsh:
Improves boomer monster immersion, note that the can opening sound effect was sourced from an actual white monster can being opened for maximum realism


lua/cinema/weapons/weapon_monster/shared.lua
```diff
@@ -22,6 +22,19 @@ function SWEP:Initialize()
     self:SetHoldType("slam")
 end
 
+function SWEP:Deploy()
+    BaseClass.Deploy(self)
+    local owner = self:GetOwner()
+
+    if IsValid(owner) then
```
swampservers: unnecessary check, remove the baseclass.deploy also

swampservers:
added dev tools access, thanks!

Kody-Quintana:
@swampservers thanks for the dev tool access, I have this "working" locally but only with `EmitSound` and not `ExtEmitSound`.

`ExtEmitSound` works with all the other sound files, just not the new one. Does `ExtEmitSound` not work with new files that aren't actually on the server yet?

Narshalbarsh:
@swampservers thanks for the dev tool access. I have this "working" locally, but only with `EmitSound` and not `ExtEmitSound`.

`ExtEmitSound` works with all the other sound files, just not the new one. Does `ExtEmitSound` not work with new files that aren't actually on the server yet?

WinterPhoenix (approved):
We already talked about this elsewhere, but `ExtEmitSound` is deprecated and only really useful for lip-syncing. Looks good to me

Narshalbarsh:
Thank you winter, I'm already feeling more immersed

swampservers:
> We already talked about this elsewhere, but `ExtEmitSound` is deprecated and only really useful for lip-syncing. Looks good to me

It also handles locations (so you dont hear everything from inside a theater) and the crouch pitch shifting. Someone is already complaining about it being audible map wide

Narshalbarsh:
> It also handles locations (so you dont hear everything from inside a theater) and the crouch pitch shifting. Someone is already complaining about it being audible map wide

Seems like `ExtEmitSound` not working was only because the sound file was new, so the live preview of the dev tools couldn't use it?

Now that the wav file is merged, it works. Here's a PR https://github.com/swampservers/contrib/pull/465/files

WinterPhoenix:
The `EntityEmitSound` hook/network handling we have is supposed to deal with that regardless.

swampservers:
> The `EntityEmitSound` hook/network handling we have is supposed to deal with that regardless.

No, that is only for global volume control. The problem with EmitSound hooks is we don't know what the sound came from beyond it being a player. The ExtEmitSound makes it so when you are in a theater you can't hear soundplayer sweps from outside the theater regardless of your volume settings. Also the crouch pitch shifting is included in there.

PR #465, opened 2024-10-25
Fix mapwide sounds for monster

Narshalbarsh:
`ExtEmitSound` didn't work until the sound file was merged, but it works now.


WinterPhoenix (approved):
I'll merge this for now just to deal with it quickly. `ExtEmitSound` shouldn't need to exist, it should all be handled by the `EntityEmitSound` networking stuff already

PR #476, opened 2024-12-02
Fix projector bug caused by snow

Narshalbarsh:
This texture is no longer on the map, setting the texture to snow causes projector screen mobile theaters to not display correctly for some reason


WinterPhoenix:
Already fixed, but thanks!

